PR #1, opened 2015-11-27
Better `copyAssets` 

grabbou:
It now detects file type and handles fonts differently than the rest (well, for now it does nothing apart from that but can be easily patched in the future)

For Android, it just copies all the fonts to `targetPath/fonts` folder preserving the original name.

For iOS, it creates group reference to original fonts directory.


grabbou:
> "@Kureev LGTM"

PR #3, opened 2015-11-29
(feature) header search paths & unit tests

grabbou:
This pull requests adds a couple of things:
- Adds header search paths to projects that have headers inside
- Unit tests newly added method & prepares structure for further tests
- Fixes small issue with `assets.length` - We need to check if there are any assets since that array is always defined (Regression)


grabbou:
CC: @Kureev

src/ios/registerNativeModule.js
```diff
@@ -50,6 +50,36 @@ const addProjectToLibraries = (libraries, file) => {
...
+const addToHeaderSearchPaths = (project, path) => {
+  const config = project.pbxXCBuildConfigurationSection();
+
+  Object
+    .keys(config)
+    .filter(ref => ref.indexOf('_comment') === -1)
+    .forEach(ref => {
+      const buildSettings = config[ref].buildSettings;
+      const shouldAddSearchPath = (buildSettings.HEADER_SEARCH_PATHS || [])
+        .filter(path => path.indexOf('react-native/React/**'))
```
grabbou: We do not need to add `>= 0` or something since path never starts with `react-native`.

src/ios/registerNativeModule.js
```diff
@@ -50,6 +54,74 @@ const addProjectToLibraries = (libraries, file) => {
...
+  }
+
+  const directory = directories.reduce((topMostDir, currentDir) => {
+    return currentDir.split(path.sep) < topMostDir.split(path.sep)
+      ? currentDir
+      : topMostDir;
+  });
+
+  return `"$(SRCROOT)/${path.relative(from, directory)}/**"`;
+};
```
grabbou: Thanks to that method, it now properly handles react-native-code-push by returning:
`"$(SRCROOT)/../node_modules/react-native-code-push"` for it and my fake library that has nested `src` folder (common-case) by returning:
`"$(SRCROOT)/../node_modules/react-native-linear-gradient/src/**",`
because there are src/module/module.h and src/moduleb/moduleb.h

grabbou: For `react-native-quick-actions` it returns:
```"$(SRCROOT)/../node_modules/react-native-quick-actions/RNQuickAction/RNQuickAction"```
See: https://github.com/jordanbyron/react-native-quick-actions/tree/master/RNQuickAction/RNQuickAction

**Comment by grabbou** [comment]
To be rebased, leaving commits so it's easier to follow what was the issue.

File: src/ios/registerNativeModule.js [hunk 3]
```diff
@@ -50,6 +54,73 @@ const addProjectToLibraries = (libraries, file) => {
...
+  }
+
+  const directory = directories.reduce((topMostDir, currentDir) => {
+    return currentDir.split(path.sep) < topMostDir.split(path.sep)
+      ? currentDir
+      : topMostDir;
+  });
+
+  return `"$(SRCROOT)/${path.relative(sourceDir, directory)}/**"`;
+};
```
grabbou: This is needed so for code-push it returns `../node-modules/react-native-code-push` (not recursive because all is in top-level folder), for react-native-quick-actions it returns `../node_modules/react-native-quick-actions/RNQuickActions/RNQuickActions` (all in same folder as well), but for library that has multiple folders (just in case), e.g.:
```
- BVLinearGradient.h
- typicalNestedFolder
-- BVLinearGradientManager.h
```
it will return `.../node_modules/react-native-linear-gradient/**`

src/ios/registerNativeModule.js
```diff
@@ -50,6 +54,73 @@ const addProjectToLibraries = (libraries, file) => {
...
+ * inside it, ignoring node_modules and examples
+ */
+const getHeadersInFolder = (folder) =>
+  glob
+    .sync('**/*.h', {
+      cwd: folder,
+      nodir: true,
+      ignore: GLOB_EXCLUDE_PATTERN,
+    })
+    .map(file => path.join(folder, file));
```
grabbou: we map to absolute path so we don't have to pass `destinationFolder` to `getHeaderSearchPath` because it's already a part of `header` path. (see method below)

src/ios/registerNativeModule.js
```diff
@@ -50,6 +54,73 @@ const addProjectToLibraries = (libraries, file) => {
...
+ * To workaround that issue and make it more bullet-proof for different names,
+ * we iterate over all configurations and look if React is already there. If it is,
+ * we assume user will need newly added dependency headers in it as well.
+ */
+const addToHeaderSearchPaths = (project, path) => {
+  const config = project.pbxXCBuildConfigurationSection();
+
+  Object
+    .keys(config)
+    .filter(ref => ref.indexOf('_comment') === -1)
```
grabbou: node-xcode stores _comment for targets that describe whether its Debug or Release. We do not need that information and that's also what they do when iterating.

src/ios/registerNativeModule.js
```diff
@@ -78,6 +149,14 @@ module.exports = function registerNativeModuleIOS(dependencyConfig, projectConfi
     });
   });
 
+  const headers = getHeadersInFolder(dependencyConfig.folder);
+  if (headers.length > 0) {
```
grabbou: No action is taken when library exposes purely JS solution (e.g. react-native-side-menu)

src/ios/registerNativeModule.js
```diff
@@ -50,6 +54,80 @@ const addProjectToLibraries = (libraries, file) => {
...
+
+  if (directories.length === 1) {
+    return `"$(SRCROOT)/${path.relative(sourceDir, directories[0])}"`;
+  }
+
+  const directory = directories.reduce((topMostDir, currentDir) => {
+    const currentFolders = currentDir.split(path.sep);
+    const topMostFolders = topMostDir.split(path.sep);
+
+    if (currentFolders.length === topMostFolders.length) {
```
grabbou: That case is for a project that has:
```
src/Base/A.h
src/Layout/B.h
```
(which is react-native case).

In that case, the `src` folder will be returned.

src/ios/registerNativeModule.js
grabbou: I will add some unit tests to that method tomorrow, so let's hold with the merge & release before that happens.

src/ios/registerNativeModule.js
```diff
@@ -50,6 +54,80 @@ const addProjectToLibraries = (libraries, file) => {
...
+      return currentFolders.slice(0, -1).join(path.sep);
+    }
+
+    return currentFolders.length < topMostFolders.length
+      ? currentDir
+      : topMostDir;
+  });
```
grabbou: Otherwise, in case:
````
src/Base/A.h
src/B.h
```
arrays will be:
['users', 'grabbou', 'project', 'node_modules', 'react-module', 'src', 'base']
['users', 'grabbou', 'project', 'node_modules', 'react-module', 'src'];

We return the one that's shorter.

src/ios/registerNativeModule.js
```diff
@@ -50,6 +54,80 @@ const addProjectToLibraries = (libraries, file) => {
...
+ * it finds the top-most one and returns recursive version of it
+ */
+const getHeaderSearchPath = (sourceDir, headers) => {
+  const directories = union(
+    headers.map(path.dirname)
+  );
+
+  if (directories.length === 1) {
+    return `"$(SRCROOT)/${path.relative(sourceDir, directories[0])}"`;
+  }
```
grabbou: If all files were
```
src/A.h
src/B.h
```
unique array of paths is `users/grabbou/project/node_modules/react-module/src` and there's no extra action to be taken.

src/ios/registerNativeModule.js
```diff
@@ -50,6 +59,83 @@ const addProjectToLibraries = (libraries, file) => {
...
+
+  const directory = directories.reduce((topMostDir, currentDir) => {
+    const currentFolders = currentDir.split(path.sep);
+    const topMostFolders = topMostDir.split(path.sep);
+
+    if (currentFolders.length === topMostFolders.length
+      && lastItem(currentFolders) !== lastItem(topMostFolders)) {
+      return currentFolders.slice(0, -1).join(path.sep);
```
grabbou: If files are on the same level but from different folders, e.g.:
```
src/Base/A.h
src/Layout/B.h
```
the array will be:
['users', 'grabbou', 'project', 'node_modules', 'react-module', 'src', 'base']
['users', 'grabbou', 'project', 'node_modules', 'react-module', 'src', 'layout'];

Since length is the same (same level) but folders are different, we return src. If folders are the same, there's no need to return outer folder hence that check.

grabbou:
This is to be merged after unit tests are added so we have 100% confidence that these methods do work and we can refactor their naming & structure later. From my internal testing it looks like all edge-cases are handled, but needs confirmation.

Kureev:
Well, what can I do for it? You want me to test ios? Which libs?

grabbou:
Just the style tips and general look and feel :) I have unit test in progress that will cover all that.

README.md
```diff
@@ -1,7 +1,7 @@
-rnpm link
+rnpm link [![Circle CI](https://img.shields.io/circleci/project/rnpm/rnpm-plugin-link/master.svg)](https://circleci.com/gh/rnpm/rnpm-plugin-link)
```
Kureev: Yoohoo, we have a CircleCI badge! :dancer:

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "1.1.1",
   "description": "rnpm plugin that links native dependencies to your project",
   "main": "index.js",
+  "scripts": {
+    "test": "eslint ./ && mocha ./test --recursive"
```
Kureev: Well, I have something to add here. I insist to use local dependencies over global ones. I think I don't need to explain "why?". So I propose to change it to something like this: 
```bash
node_modules/eslint/bin/eslint . && ./node_modules/mocha/bin/mocha
```
(./test --recursive is a default settings afaik)

package.json
```diff
@@ -36,6 +41,8 @@
   },
   "devDependencies": {
```
Kureev: According to proposed changes above, we need to add `eslint` and `mocha as a dev deps here

src/action.js
```diff
@@ -50,12 +50,12 @@ module.exports = function link(config, args) {
         registerDependencyIOS(dependencyConfig.ios, project.ios);
       }
 
-      if (project.android && dependencyConfig.assets) {
+      if (project.android && dependencyConfig.assets.length > 0) {
```
Kureev: mb it'll be more elegant to write `!emtpy(dependencyConfig.assets)`, but feel free not to do it, it's just a notice

src/groupFilesByType.js
```diff
@@ -11,6 +11,6 @@ mime.define({
   'font/truetype': ['ttf'],
 });
 
-exports.groupByType = function groupByType(assets) {
+module.exports = function groupFilesByType(assets) {
```
Kureev: Comments missing

src/ios/getHeaderSearchPath.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+const lastItem = (array) => array[array.length - 1];
+
+/**
+ * Given an array of headers it returns search path so Xcode can find them
+ * If there are multiple header files located accross different folders,
+ * it finds the top-most one and returns recursive version of it
+ */
+module.exports = function getHeaderSearchPath(sourceDir, headers) {
+  const directories = union(
```
Kureev: not sure, but I think it may fit one line < 80 chars

src/ios/getHeaderSearchPath.js
```diff
@@ -0,0 +1,38 @@
...
+    headers.map(path.dirname)
+  );
+
+  if (directories.length === 1) {
+    return `"$(SRCROOT)/${path.relative(sourceDir, directories[0])}"`;
```
Kureev: if `directories[0]` is a `targetDir`, I'd like to add a line above declaring it. Assumptions in the code is always bad.

src/ios/registerNativeModule.js
```diff
@@ -78,6 +81,14 @@ module.exports = function registerNativeModuleIOS(dependencyConfig, projectConfi
     });
   });
 
+  const headers = getHeadersInFolder(dependencyConfig.folder);
+  if (headers.length > 0) {
```
Kureev: we can use the same function `empty` as I proposed above:
```js
if (!emtpy(headers)) {
```
imo ^ makes code easier to read. Also not pushing this change,

test/groupFilesByType.spec.js
```diff
@@ -0,0 +1,21 @@
...
+      'fonts/a.ttf',
+      'fonts/b.ttf',
+      'images/a.jpg',
+      'images/c.jpeg',
+    ];
+
+    const groupedFiles = groupFilesByType(files);
+
+    expect(groupedFiles.font).to.contain('fonts/a.ttf');
+    expect(groupedFiles.image).to.contain('images/a.jpg');
```
Kureev: what about `b.ttf` and `c.jpeg`? I think it's worth to include them into test.

src/ios/getHeaderSearchPath.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  const directory = directories.reduce((topMostDir, currentDir) => {
```
Kureev: `topMostDir` => `topDir`?

src/ios/getHeaderSearchPath.js
```diff
@@ -0,0 +1,38 @@
...
+ * Returns last item from an array
+ */
+const lastItem = (array) => array[array.length - 1];
+
+/**
+ * Given an array of headers it returns search path so Xcode can find them
+ * If there are multiple header files located accross different folders,
+ * it finds the top-most one and returns recursive version of it
+ */
+module.exports = function getHeaderSearchPath(sourceDir, headers) {
```
Kureev: Probably it may be divided for two functions: search for a directory inside `headers` argument (it seems name `files` sounds more valuable for this argument) and finding common directory for all these `files` (`headers`)

src/ios/getHeaderSearchPath.js
```diff
@@ -0,0 +1,38 @@
...
+  if (directories.length === 1) {
+    return `"$(SRCROOT)/${path.relative(sourceDir, directories[0])}"`;
+  }
+
+  const directory = directories.reduce((topMostDir, currentDir) => {
+    const currentFolders = currentDir.split(path.sep);
+    const topMostFolders = topMostDir.split(path.sep);
+
+    if (currentFolders.length === topMostFolders.length
+      && lastItem(currentFolders) !== lastItem(topMostFolders)) {
```
Kureev: we need some kind of agreement about where to put `&&`. You want to put them on a new line?

package.json
grabbou: when referenced from inside npm scripts, `node_modules/bin` is added to PATH so these are actually called from inside node_modules (local stuff).

package.json
grabbou: They are already there https://github.com/rnpm/rnpm-plugin-link/blob/feature/fix-header-search-paths/package.json#L45-L46

src/action.js
grabbou: We can do, but is it worth extra function?

src/groupFilesByType.js
grabbou: Yeah, I'll add them in.

test/groupFilesByType.spec.js
grabbou: Yeah, we can compare the whole array.

src/ios/getHeaderSearchPath.js
grabbou: Yep, or outerDir?

src/ios/getHeaderSearchPath.js
grabbou: It doesn't matter, but since I usually write
```js
return a == b
  ? a
  : b
```
that's quite intuitive. What Airbnb says about that? :dancer:

src/ios/getHeaderSearchPath.js
grabbou: We can extract that. I'll play around.

README.md
grabbou: :D TravisCI was not working for some reason and seems like CircleCI is faster :dancer:

src/ios/getHeaderSearchPath.js
grabbou: Yep, it's just personal preference in this case.

README.md
Kureev: lol

package.json
Kureev: Cool, didn't knew this

package.json
Kureev: :+1:

src/action.js
Kureev: It's extra code for sure, but readability is better. Well, I'll keep it up to you, `.length > 0` is also fine for me.

src/ios/getHeaderSearchPath.js
Kureev: ok, no probs

src/ios/getHeaderSearchPath.js
Kureev: whatever, topMostDir just sounds weird :)

PR #4, opened 2015-12-01
Comments, micro-optimizations etc

Kureev:


src/action.js
```diff
@@ -6,59 +6,62 @@ const registerDependencyAndroid = require('./android/registerNativeModule');
...
+ * Namespace log messages
+ */
 log.heading = 'rnpm-link';
 
 /**
  * Returns an array of dependencies that should be linked/checked.
+ * @param {Object} pconfig Project config
  */
-const getProjectDependencies = () => {
-  const pjson = require(path.join(process.cwd(), './package.json'));
```
grabbou: That was intended. Becase the check on line 31 would've returned false if there was no package.json. Otherwise, we get require error from the very first moment. I am the fun of the previous version.

src/action.js
```diff
@@ -6,59 +6,62 @@ const registerDependencyAndroid = require('./android/registerNativeModule');
 const registerDependencyIOS = require('./ios/registerNativeModule');
 const copyAssetsAndroid = require('./android/copyAssets');
 const copyAssetsIOS = require('./ios/copyAssets');
+const pjson = require(path.join(process.cwd(), './package.json'));
```
grabbou: As I said, will throw from the very first moment providing bad experience.

Kureev: Got it. Will change it soon

grabbou: To be more explicit, you can add that require above line 37 so we do not require from inside a function. On the other hand, we do not need that `pjson` anywhere in the action, so might be good leaving it inside that function.

src/action.js
```diff
@@ -6,59 +6,62 @@ const registerDependencyAndroid = require('./android/registerNativeModule');
...
  */
 module.exports = function link(config, args) {
-  const project = config.getProjectConfig();
+  const pconfig = config.getProjectConfig();
 
-  if (!project) {
+  if (!pconfig) {
     log.error('ERRPACKAGEJSON', `No package found. Are you sure it's a React Native project?`);
     return;
   }
```
grabbou: Below that line all calls to process.cwd(), package.json can be considered `safe`.

src/action.js
Kureev: Ideally, I'd like to re-use `getConfig` here. I'll think about it

grabbou: I don't think that's needed. I actually do not see the point of having it since it just proxies to `require`. And actually inside `getConfig` there's even no need to check for `fs.existsSync` since require will throw anyway, so you can wrap that instead.

Kureev: Yeah, good point

src/action.js
```diff
@@ -6,59 +6,62 @@ const registerDependencyAndroid = require('./android/registerNativeModule');
...
 
   const packageName = args[0];
-  const dependencies = packageName ? [packageName] : getProjectDependencies();
+  const dependencies = packageName ? [packageName] : getProjectDependencies(pjson);
 
   dependencies
     .forEach(name => {
-      const dependencyConfig = config.getDependencyConfig(name);
```
grabbou: I am not actually sure if the naming changes are improving readability. I would've left the `project` and rename `dependencyConfig` to `dependency`.

src/action.js
```diff
@@ -6,59 +6,62 @@ const registerDependencyAndroid = require('./android/registerNativeModule');
...
 
 /**
  * Returns an array of dependencies that should be linked/checked.
+ * @param {Object} pconfig Project config
  */
-const getProjectDependencies = () => {
-  const pjson = require(path.join(process.cwd(), './package.json'));
-  return Object.keys(pjson.dependencies).filter(name => name !== 'react-native');
-};
+const getProjectDependencies = (pconfig) =>
```
grabbou: This is quite misleading because in below method `pconfig` refers to result from `getProjectConfig` but we receive `pjson` here.

PR #5, opened 2015-12-05
Cut $(SRCROOT) from pblistPath

ptmt:
It seems Cocoapods adds this to pbxProject (unconfirmed) and this line is breaking plist parsing.


grabbou:
Thanks! That might be actually the case. Anyway, since we join the path with `SRCROOT` anyway, replacing that (if it exists) sounds like a thing worth doing.

Issue #6, opened 2015-12-05
Suggestion: Moving some common methods to core for reusability


ptmt:
Would be nice to have some well-tested common methods somewhere in utils. This could reduce dependencies required by the plugin code. I'd like to have these:

`getPlist()` — to obtain parsed plist. I used the same code https://github.com/rnpm/rnpm-plugin-link/blob/master/src/ios/copyAssets.js#L13 here: https://github.com/ptmt/rnpm-plugin-apphub/blob/master/src/projectMetadata.js (potentially it's useful for `rnpm-plugin-build` plugin too).

`GLOB_EXCLUDE_PATTERN` or even `searchFile(template)` for searching files in project source directory. 

Related: https://github.com/rnpm/rnpm-plugin-link/pull/5

grabbou:
Woah, I have to say your plugin looks really exciting! Yep, I do agree and the number of utils is only going to increase as long as we add in new things.

I think that the most reasonable approach would be to set up `rnpm-utils-ios` (just like `Cocoapods/xcodeproj` library) with all these methods extracted/defined.

The current state of `rnpm-plugin-link` is in a stable enough stage to be extracted, so if that's something that'd worked for you - let me know and we can do that.

ptmt:
Another approach would be extending function `config` argument, like `config.getPlist() : ParsedPlist` or `config.getXcodeproj() : Pbxproj`, etc. 

Ok, let's see what code would be shared by another plugins, and extract it if needed.

grabbou:
Do you find any other methods useful? I am thinking about publishing `rnpm-ios` with all files from `ios` folder except `register/unregisterNativeModule` and `link/unlinkAssets`.

grabbou:
Let's extract that to `rnpmjs` when #45 is merged so we can also use uniq/union etc. in rnpm itself.

grabbou:
I think for now you should be able to `require('rnpm-plugin-link/src/ios/<func>')`. We will extract it later (or expose properly), but since we are transitioning to react-native, we can't extract them at this point.

PR #8, opened 2015-12-05
Create `Resources` group [ios]

grabbou:
This pull request implements automatic `resources` group creation instead of printing fatal warning when copying assets.


grabbou:
@Kureev let me know wdyt, some unit tests to be done later today/tmrw.

src/ios/addPBXGroup.js
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * This is done by accessing `PBXProjectSection` that
+ * by the `xcode` library design returns an object with keys:
+ * `uuid` & `uuid_comment`. We only access the first key in the array
+ * (just like `xcode` internally) and return that group details from `PBXGroup`
+ * section.
+ */
+const getMainGroup = (project) => {
+  const projectSection = project.pbxProjectSection();
+  const projectId = Object.keys(projectSection)[0];
```
Kureev: Is `[0]` always a projectId?

src/ios/getHeadersInFolder.js
```diff
@@ -1,7 +1,7 @@
 const glob = require('glob');
 const path = require('path');
 
-const GLOB_EXCLUDE_PATTERN = ['node_modules/**', 'Examples/**', 'examples/**'];
+const GLOB_EXCLUDE_PATTERN = ['node_modules/**', 'Pods/**', 'Examples/**', 'examples/**'];
```
Kureev: I think we already have this one form @ptmt, right?

grabbou: Kind of, he added that to `rnpm` itself that has similar GLOB. the difference is that in rnpm we look for .xcodeproj, here we look for headers so I just thought like it might be useful.

Kureev: Make sense

grabbou: Further discussion in #6 (this kind of stuff could go there as well, as per @ptmt suggestion `findFiles(type, folder)` function

src/ios/addPBXGroup.js
grabbou: Yep. That's how `xcode` accesses that and they have that unit tested.

grabbou: I am not comfortable with that statement to be fair, so I'd consider better solution to that.

Kureev:
What about tests?

grabbou:
screen shot 2015-12-05 at 21 25 20
In progress, but it's pretty straight forward. I am delivering all iOS stuff on another branch.

grabbou:
@Kureev - implementation is now way more smarter (80% smaller)

Kureev:
Well, I'd prefer ship tests with a feature. So, mb you consider to add them here?

Kureev:
(besides tests it looks good to me and you can merge and :ship: it)

grabbou:
Tests were added.

PR #9, opened 2015-12-07
More iOS unit tests

grabbou:
That pull request is meant to be merged before the one from create-pbx-group since it adds in more unit tests to iOS methods.

It also adds unit tests for `action.js` that's why it's important to get it merged before #10 so I can rebase and make sure it passes.

All methods are tested on `react-native-linear-gradient`.


grabbou:
Everything tested except `copyAssets` and `registerNativeModule` (so-called `main` methods). Can be merged.

test/action.spec.js
```diff
@@ -0,0 +1,112 @@
+const chai = require('chai');
+const expect = chai.expect;
+const sinon = require('sinon');
+const mock = require('mock-require');
+const log = require('npmlog');
+const path = require('path');
+
+const action = (config, args) => require('../src/action')(config, args || []);
```
grabbou: We have to do it because of line 15.

test/action.spec.js
```diff
@@ -0,0 +1,112 @@
...
+
+log.level = 'silent';
+
+describe('action', () => {
+
+  beforeEach(() => {
+    delete require.cache[require.resolve('../src/action')];
```
grabbou: Since require puts all the things into `cache`, we have to invalidate it before each suite runs so that mocks are working. In `rnpm`, it works w/o that because we require inside functions (during runtime). In this case, we have top-level requires for e.g. `registerNativeModuleIOS` and that had to be handled. We can move that to `utils` once we find it reusable.

test/action.spec.js
```diff
@@ -0,0 +1,112 @@
...
+    const spy = sinon.spy(log, 'error');
+
+    const config = {
+      getProjectConfig: () => null,
+    };
+
+    action(config);
+
+    expect(spy.calledWith('ERRPACKAGEJSON')).to.be.true;
+  });
```
grabbou: That test suite tests older `rnpm-link` behaviour (where we used to return null). Has to be changed after being merged to master because another PR `sync-with-rnpm-changes` has been merged in the meantime. It's one-liner so I'll take care of that once you say `LGTM`.

src/ios/getProducts.js
```diff
@@ -0,0 +1,12 @@
...
+ * Given xcodeproj it returns list of products ending with
+ * .a extension, so that we know what elements add to target
+ * project static library
+ */
+module.exports = function getProducts(project) {
+  return project
+    .pbxGroupByName('Products')
+    .children
+    .map(c => c.comment)
+    .filter(c => c.indexOf('.a') > -1);
```
grabbou: Nothing changed here, just moved out to separate file.

test/action.spec.js
Kureev: action === command?

Kureev: since we have a possibility to add multiple commands inside one plugin, I'd compose a new PR to replace src/action by src/link

test/action.spec.js
Kureev: can we mock these require calls instead and use mock.stopAll()?

test/action.spec.js
Kureev: You're testing if you have an error log if action fails? I think this one can go or at least reworked.

grabbou: There's no other way https://github.com/rnpm/rnpm-plugin-link/blob/master/src/action.js#L27-L32

test/action.spec.js
grabbou: We are changing the mocks on per suite basis, which means we have to do it each time (actually there are no cons imo)

test/action.spec.js
grabbou: Yes, but I am keeping the naming with current master. You can submit a PR but maybe let's wait till these two get merged and ideally #10 so that we are clean?

test/action.spec.js
Kureev: I'm not saying that there are any cons here. I have some reasons to do it in a proposed way: 
- It makes tests simpler to read and understand. In my opinion mocking something is more obvious than manual require cache cleaning
- In a case of consistency; we're using mock-require library anyway, so I'd like to stick to one approach mocking require calls

I'm fine to merge it as it is, but then I'll compose a PR to change it to mock-require. Deal?

test/action.spec.js
Kureev: What about not catching it and let our program to fail with this error instead of catching it and doing the same?

Kureev:
I found no critical issues here. We can merge it right now, but I'd prefer to clean up some comments first

test/action.spec.js
grabbou: Because we want to print nice message instead of a stack trace with `cannot require .../package.json` that is not that nice :P Anyway, this is again out of scope for this PR, if we want to change that behaviour - we can, but this just adds tests to it.

test/action.spec.js
grabbou: Have you seen line `4` -> `const mock = require('mock-require');` ? We are doing all the things in the same way. It's just that since require calls in `action.js` are top-level ones and cached after module is required for the first time, I clean up that cache in `beforeEach` so requireMock can set them up (require mock cannot mock anything if it's been already required and put into cache).

grabbou: Quick example:
```js
mock('../src/ios/registerNativeModule', noop)
require('./src/actions.js)() // this gets `noop` mocked, OK

mock('../src/ios/registerNativeModule', noop2)
require('./src/actions.js)() // this gets `noop` instead of noop2, BAD
```

Kureev: Ok, so just change `../src/action` to `path.join(__dirname, '..', 'src', 'action')` and we're fine here

test/action.spec.js
Kureev: ok

test/action.spec.js
grabbou: Why? Require does not need to have it

PR #11, opened 2015-12-08
Merge tests back

grabbou:
Because #9 was into #8  and #8 was merged first.

CC: @Kureev


Kureev:
That's why we need to have atomic PRs

grabbou:
They were atomic. But #9 was using test suite invented in #8 and I wanted to add in more tests in the meantime.

Kureev:
> But #9 was using test suite invented in #8 and I wanted to add in more tests in the meantime.

That means you have to cherry-pick commit with this test if you so want to

PR #15, opened 2015-12-30
Fix the path seperator when searching for iOS header files

axemclion:
Tests fail on Windows since an incorrect path separator was used. Instead of hard coding `\`, this PR basically adds `path.sep`, thus adding the right path. This way, iOS projects can be linked from Windows also.


Kureev:
Wow, interesting. I thought there's no way to run iOS projects on Windows. Anyway, PR lgtm.

axemclion:
You actually cannot run the project on Windows, but this is for a case when you want to link the libraries, and then build using something like a cloud build server.

Kureev:
Ah, nice one, really :+1:

PR #16, opened 2015-12-30
Use a dependency's sourceDir when writing to settings.gradle

axemclion:
Currently, it hardcodes `/android` as the path even if a dependency
specifies a sourceDir. When a dependency specifies a sourceDir, that
should be respected and added to settings.gradle


axemclion:
This PR is based on the discussion we had on Discord

Kureev:
Thanks, @axemclion <3

PR #17, opened 2016-01-09
Support "unlink" functionality for iOS

grabbou:
Todo: 
- [X] remove project from Xcode
- [X] unit tests


src/ios/unlinkAssets.js
```diff
@@ -0,0 +1,58 @@
...
+      `Group 'Resources' does not exist in your XCode project. There is nothing to unlink.`
+    );
+  }
+
+  const plist = plistParser.parse(
+    fs.readFileSync(plistPath, 'utf-8')
+  );
+
+  const fonts = (assets.font || [])
+    .map(asset =>
```
Kureev: mb `forEach`?

Kureev:
Looks good so far! When you're planning to finish it?

grabbou:
I hope today!

grabbou:
Fixed eta: End of today

Kureev:
new ETA? :laughing:

grabbou:
I am having exams this week so going slower than I thought, but it's almost there :dancer:

Kureev:
Can you rebase this branch?

grabbou:
Will do right after I send over test suites :)

grabbou:
Ready for review.

src/ios/addToHeaderSearchPaths.js
```diff
@@ -1,29 +1,5 @@
...
-        .filter(path => path.indexOf('react-native/React/**'))
-        .length > 0;
+const mapHeaderSearchPaths = require('./mapHeaderSearchPaths');
 
-      if (shouldAddSearchPath) {
-        buildSettings.HEADER_SEARCH_PATHS.push(path);
-      }
-    });
+module.exports = function addToHeaderSearchPaths(project, path) {
+  mapHeaderSearchPaths(project, searchPaths => searchPaths.concat(path));
```
Kureev: Usually you return an array from the `map`-like functions

grabbou: Usually you pass an array to map, here we pass project and mapping happens inside it (check the `func` application. It's mutating because of underyling library design, so the only way to get over that was to keep that in a separate stuff.

src/link.js
```diff
@@ -21,6 +21,50 @@ const getProjectDependencies = () => {
...
+    log.info(`Linking ${dependency.name} android dependency`);
+
+    const didLink = registerDependencyAndroid(
+      dependency.name,
+      dependency.config.android,
+      project.android
+    );
+
+    if (didLink) {
+      log.info(`Android module ${packageName} has been successfully linked`);
```
Kureev: if you notify a user in a happy scenario, you should do the same for the opposite one.

src/link.js
```diff
@@ -21,6 +21,50 @@ const getProjectDependencies = () => {
...
+    }
+  }
+
+  if (project.ios && dependency.config.ios) {
+    log.info(`Linking ${dependency.name} ios dependency`);
+
+    if (registerDependencyIOS(dependency.config.ios, project.ios)) {
+      log.info(`iOS module ${packageName} has been successfully linked`);
```
Kureev: Same as above ^

src/link.js
```diff
@@ -21,6 +21,50 @@ const getProjectDependencies = () => {
...
+  }
+
+  if (project.ios) {
+    log.info('Linking assets to ios project');
+    copyAssetsIOS(assets, project.ios);
+  }
+
+  if (project.android) {
+    log.info('Linking assets to android project');
+    copyAssetsAndroid(assets, project.android.assetsPath);
```
Kureev: If we introduce notifications about linking status, why don't we have them for the assets?

src/ios/addToHeaderSearchPaths.js
Kureev: Mmmm, based on what you've said, I see a two options which can help us to move forward with this:
- Rename it to something like `patchHeaderSearchPaths`
- Add a comment that this function is impure and it mutates some state inside

Personally I prefer the first option, but I'll be fine with the second one also.

Kureev:
Fix user notifications to make them consistent and we're ready to go :rocket:

src/link.js
grabbou: I was just following `unlink` part (trying to match the style). Any changes e.g. `Android module already linked` should be subject to another PR.

grabbou:
They are consistent :P But meh, I'll update both files anyway.

Kureev:
I mean add them for failed linking and assets also :smiley:

PR #18, opened 2016-01-11
Hooks

Kureev:
## New functionality
Introduced a `before` and `after` hooks.

## Usage
```json
"rnpm": {
  "hooks": {
    "before": "./bin/run-something",
    "after": "./bin/run-something-else"
  }
}
```

### DIFF
--- a/src/link.js
+++ b/src/link.js
@@ -12,6 +12,27 @@ const copyAssetsIOS = require('./ios/copyAssets');
 
 log.heading = 'rnpm-link';
 
+function makeHook(dependency, name) {
+  return (cb) => {
+    if (dependency.config.hooks && dependency.config.hooks[name]) {
+      const hook = spawn(dependency.config.hooks[name], {
+        stdio: 'inherit',
+        stdin: 'inherit',
+      });
+
+      hook.on('close', function prelink(code) {
+        if (code) {
+          process.exit(code);
+        }
+
+        cb();
+      });
+    } else {
+      cb();
+    }
+  };
+}
+
 /**
  * Returns an array of dependencies that should be linked/checked.
  */
@@ -62,43 +83,8 @@ module.exports = function link(config, args, callback) {
   );
 
   const tasks = dependencies.map((dependency) => (next) => {
-    const prelink = (cb) => {
-      if (dependency.config.hooks && dependency.config.hooks.prelink) {
-        const hook = spawn(dependency.config.hooks.prelink, {
-          stdio: 'inherit',
-          stdin: 'inherit',
-        });
-
-        hook.on('close', function prelink(code) {
-          if (code) {
-            process.exit(code);
-          }
-
-          cb();
-        });
-      } else {
-        cb();
-      }
-    };
-
-    const postlink = (cb) => {
-      if (dependency.config.hooks && dependency.config.hooks.postlink) {
-        const hook = spawn(dependency.config.hooks.postlink, {
-          stdio: 'inherit',
-          stdin: 'inherit',
-        });
-
-        hook.on('close', function postlink(code) {
-          if (code) {
-            process.exit(code);
-          }
-
-          cb();
-        });
-      } else {
-        cb();
-      }
-    };
+    const prelink = makeHook(dependency, 'prelink');
+    const postlink = makeHook(dependency, 'postlink');
 
     async.waterfall([prelink, (cb) => {
       if (project.android && dependency.config.android) {
--- a/src/link.js
+++ b/src/link.js
@@ -25,7 +25,7 @@ const getProjectDependencies = () => {
  *
  * If optional argument [packageName] is provided, it's the only one that's checked
  */
-module.exports = function link(config, args) {
+module.exports = function link(config, args, callback) {
 
   try {
     const project = config.getProjectConfig();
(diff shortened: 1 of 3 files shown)


Kureev:
Ready to be reviewed! I'll compose a separated PR in rnpm repo itself to make a README patch

src/link.js
```diff
@@ -75,17 +61,69 @@ module.exports = function link(config, args) {
...
 
-  if (isEmpty(assets)) {
-    return;
-  }
+  const tasks = dependencies.map((dependency) => (next) => {
+    const before = (cb) => {
+      if (dependency.config.hooks && dependency.config.hooks.before) {
+        exec(dependency.config.hooks.before, function before(err, stdout, stderr) {
+          if (!err && !stderr) {
+            console.log(stdout);
```
grabbou: console.log

src/link.js
```diff
@@ -75,17 +61,69 @@ module.exports = function link(config, args) {
...
+            cb();
+          } else {
+            throw Error(err || stderr);
```
grabbou: Do we have to throw here? Can't we just `cb(err)` ?

src/link.js
```diff
@@ -75,17 +61,69 @@ module.exports = function link(config, args) {
     asset => path.basename(asset)
   );
 
-  if (isEmpty(assets)) {
-    return;
-  }
+  const tasks = dependencies.map((dependency) => (next) => {
```
grabbou: I thought that hooks were going to happen transparently for the plugin in the `rnpm` instead of `rnpm-link`. That will require quite a lot of work to be done by the community authors. I generally like the approach but would prefer it to run in rnpm cli.

src/link.js
```diff
@@ -75,17 +61,69 @@ module.exports = function link(config, args) {
...
+    const before = (cb) => {
+      if (dependency.config.hooks && dependency.config.hooks.before) {
+        exec(dependency.config.hooks.before, function before(err, stdout, stderr) {
```
grabbou: So that's custom hooks on per-project basis specified by users?

src/link.js
grabbou: Ah, ok. So that's just for link dependencies only.

grabbou:
Quick question - do we really have to introduce async here at all? It does not matter since it's CLI, I think it would've been easier to just use https://github.com/shelljs/shelljs#execcommand--options--callback ?

src/link.js
grabbou: In my opinion, we should just stick to `npm` scripts. We could either:
```json
{
   "scripts": {
       "start": "node index.js",
       "post-rnpm-link": "abc"
   } 
}
```
so it's easier for the people to follow.

src/link.js
```diff
@@ -75,17 +61,69 @@ module.exports = function link(config, args) {
...
+      } else {
+        cb();
+      }
+    };
 
-  if (project.ios) {
-    log.info('Linking assets to ios project');
-    copyAssetsIOS(assets, project.ios);
-  }
+    const after = (cb) => {
```
grabbou: Isn't it better to write:
```js
const after = done =>  (dependency.config.hooks || {}).before 
   ? exec(dependency.config.hooks.after, done)
    : done()
};
```
I would also suggest to add in `hooks` to a function that makes config in `rnpm` automatically so that we don't have to check if `dependency.config.hooks` is defined.

Kureev:
1. Well, for me both of these variants are equivalent. I'm fine with promises also, but it's already written using `async` and callbacks. Do you think we need to rewrite it?

2. Where else we can use these hooks? Can you provide any examples? Imo `link` is the only place where we iterate dependencies.

3. > I am afraid that plugin author has to write some code snippets in order to support that feature (they can forget about it etc.

Not sure I can follow this. We're not pushing library authors to support this feature, it's up to them if it's needed.

grabbou:
> Where else we can use these hooks? Can you provide any examples? Imo link is the only place where we iterate dependencies.

Well, you have answered yourself in 3 point :D If we are aiming for the hooks to also work in other plugins and actions, I would aim for something less generic.

> Do you think we need to rewrite it?

Well, it's not a deal breaking feature which means we have some time to work on it ;) I am fine with `async`, but I can see a bit more pros of going all the way down with Promises.

grabbou:
Shouldn't we add `return callback()` here as well? https://github.com/rnpm/rnpm-plugin-link/blob/126f7768ef13a073e93297a2260cc6d94744bf23/src/link.js#L55

src/link.js
```diff
@@ -10,6 +12,25 @@ const copyAssetsIOS = require('./ios/copyAssets');
...
+    stdio: 'inherit',
+    stdin: 'inherit',
+  });
+
+  hook.on('close', function prelink(code) {
+    if (code) {
+      log.error(`Failed to link a "${dependency.name}" module`);
+    }
+
+    cb();
```
grabbou: Are you sure we want to swallow errors? NPM stops and I think it's reasonable for us to do either. Just cb(new Error())` should be enough with that `falied to link...` message instead of log.error call (it will also make the code to log from one place only)

grabbou:
@Kureev generally happy, you can merge it straight away - just check my inline comment

src/link.js
Kureev: This callback wouldn't be caught. I propose `log.error` + `process.exit(1)`.

grabbou: Why it wouldn't be caught? Can't we replace `async.series(... callback || () => {})` (the second noop) with something that actually exits? Or let's just do that in rnpm (exit and log if err from callback)

PR #19, opened 2016-01-18
Introduce prelink and postlink commands

Kureev:
As we discussed in Discord, I've made a standalone repo for hooks wrapper.
Basically, since now, users can make their own plugins with a hooks support.

## How it works
Basically, I moved `makeHook` function to the standalone repo to allow other people use it.

## Scope of this PR
Replace in-place function by the `rnpm-hooks` module. Added missing `require` for testing and adjusted link code to take care about generating hooks. I'd like to keep `rnpm-hooks` as tiny as possible (and not to depend on our `dependency` object structure), so I think [this kind of logic](https://github.com/rnpm/rnpm-plugin-link/compare/chore/better-hooks?expand=1#diff-1dbe17fda47e0a0d65b16ad4646552d0R103) should be scoped in this specific plugin.

See [rnpm-hooks](https://github.com/rnpm/rnpm-hooks) for the further reading.

### DIFF
--- a/src/link.js
+++ b/src/link.js
@@ -2,6 +2,7 @@ const spawn = require('child_process').spawn;
 const path = require('path');
 const log = require('npmlog');
 const uniq = require('lodash.uniq');
+const makeHook = require('rnpm-hooks');
 const async = require('async');
 
 const isEmpty = require('./isEmpty');
@@ -11,25 +12,6 @@ const copyAssetsAndroid = require('./android/copyAssets');
 const copyAssetsIOS = require('./ios/copyAssets');
 
 log.heading = 'rnpm-link';
-
-const makeHook = (dependency, name) => (cb) => {
-  if (!dependency.config.hooks || !dependency.config.hooks[name]) {
-    return cb();
-  }
-
-  const hook = spawn(dependency.config.hooks[name], {
-    stdio: 'inherit',
-    stdin: 'inherit',
-  });
-
-  hook.on('close', function prelink(code) {
-    if (code) {
-      cb(new Error(`Failed to link a "${dependency.name}" module`));
-    }
-
-    cb();
-  });
-};
 
 /**
  * Returns an array of dependencies that should be linked/checked.
@@ -113,11 +95,29 @@ module.exports = function link(config, args, callback) {
   };
 
   const tasks = dependencies.map((dependency) => (next) => {
-    const prelink = makeHook(dependency, 'prelink');
-    const postlink = makeHook(dependency, 'postlink');
-    const link = makeLink(dependency);
+    const queue = [];
 
-    async.waterfall([prelink, link, postlink, next]);
+    var prelink;
+    var postlink;
+
+    if (dependency.config.hooks) {
+      prelink = dependency.config.hooks.prelink;
+      postlink = dependency.config.hooks.postlink;
+    }
+
+    if (prelink) {
+      queue.push(makeHook(prelink));
+    }
+
+    queue.push(makeLink(dependency));
+
+    if (postlink) {
+      queue.push(makeHook(postlink));
+    }
+
+    queue.push(next);
+
+    async.waterfall(queue);
   });
 
   async.series(tasks, callback || () => {});
--- a/src/link.js
+++ b/src/link.js
@@ -11,6 +11,8 @@ const copyAssetsIOS = require('./ios/copyAssets');
 
 log.heading = 'rnpm-link';
 
+const commandStub = (cb) => cb();
+
 /**
  * Returns an array of dependencies that should be linked/checked.
  */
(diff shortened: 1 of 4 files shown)


grabbou:
Yeah, that works :)

Have you thought about moving async to hooks repo as well so that we just:
```js
const hooks = require('hooks')();
hooks.before([...])
hooks.after([...])
hooks.run(link, cb)
```
?

Kureev:
Can you explain how it'll benefit for us? It looks like an `async` interface analog but + state machine (why?). Or you have any cases where this hooks s/m may be handy for us?

grabbou:
Then I guess the name shouldn't be hooks but something like `func-utils` or maybe just `utils`. The whole point was to avoid writing async all over in every project that wants to use hooks, but do:
```js
hooks.run([before, command, after])`
```
instead (the previous example was quite shit, ignore it)

src/link.js
```diff
@@ -113,11 +95,29 @@ module.exports = function link(config, args, callback) {
...
-    const link = makeLink(dependency);
+    const queue = [];
+
+    var prelink;
+    var postlink;
+
+    if (dependency.config.hooks) {
+      prelink = dependency.config.hooks.prelink;
+      postlink = dependency.config.hooks.postlink;
+    }
```
grabbou: Can't we just
```
const prelink = dependency.config.hooks.prelink;
const postlink = dependency.config.hooks.postlink;
```
?

src/link.js
```diff
@@ -113,11 +95,29 @@ module.exports = function link(config, args, callback) {
...
+    queue.push(makeLink(dependency));
+
+    if (postlink) {
+      queue.push(makeHook(postlink));
+    }
+
+    queue.push(next);
 
-    async.waterfall([prelink, link, postlink, next]);
+    async.waterfall(queue);
```
grabbou: There's no `callback` to waterfall. I think it should be `async.waterfall([...], next)` instead.

src/link.js
```diff
@@ -113,11 +95,29 @@ module.exports = function link(config, args, callback) {
...
+    var prelink;
+    var postlink;
+
+    if (dependency.config.hooks) {
+      prelink = dependency.config.hooks.prelink;
+      postlink = dependency.config.hooks.postlink;
+    }
+
+    if (prelink) {
+      queue.push(makeHook(prelink));
```
grabbou: as I said in the comments, maybe something with `command` name would've been better

src/link.js
Kureev: if we don't have any hooks, the code above will fail (Can't get prelink of undefined). That's why I wrapped it into check.

src/link.js
Kureev: I totally agree with it. That was a point I wanted to discuss with you yesterday when you went offline :+1:

src/link.js
Kureev: Oh shit, true!

Kureev:
so something like this:
```js
const makeCommand = require('rnpm-commands').makeCommand;
const run = require('rnpm-commands').run;

const someCommand = makeCommand('some-shell-command-here');
const pluginFunction = () => {...};

run([someCommand, pluginFunction]);
```

But in this case we still need to have an `async.series` to run it. So I'd prefer to stick to one `makeCommand` for now. If people will request a `run` command or something like it - we can add it at any time (but I don't think it'll value to them).

Kureev:
> Maybe it could have two functions - wrapShellCommand
> ```js
> const before = beforeLink.map(wrapShellCommand);
> ```
> to return an array of functions with spawn we can execute and - run

Can you explain how we may have a list of commands here? One string always maps to one command. There's no point to have an array here. Do I miss something?

Kureev:
> I guess that once demand is higher, map(wrapShellCommand) can be moved here 
> https://github.com/rnpm/rnpm/blob/master/src/config/index.js#L22 and
> https://github.com/rnpm/rnpm/blob/master/src/config/index.js#L38
> so you can just run them w/o thinking about wrapping them.

So you want to wrap them automatically? Well, sounds good. Indeed, we can wrap all custom commands even before we run a plugin :+1: I'll take care about it!

grabbou:
> Can you explain how we may have a list of commands here? One string always maps to one command. There's no point to have an array here. Do I miss something?

My bad, for some reason I assumed it's an array :dancer:

Kureev:
@grabbou can you review it? :8ball:

src/link.js
```diff
@@ -113,11 +93,27 @@ module.exports = function link(config, args, callback) {
...
-    const link = makeLink(dependency);
+    const queue = [];
+
+    var prelink;
+    var postlink;
+
+    if (dependency.config.commands) {
+      prelink = dependency.config.commands.prelink;
+      postlink = dependency.config.commands.postlink;
+    }
```
grabbou: What happens when these are not defined? Is `asycn.waterfall` capable of handling undefined `prelink`, `postlink`? I think commands are **always** defined so we should check for presence of prelink/postlink instead

Kureev: yes, if it's undefined - it'll skip it.

grabbou: My suggestion for these lines is just:
```js
async.series(
   dependencies.map(dependency => next =>
      async.waterfall([
         dependency.commands.preLink,
         makeLink(...),
         dependency.commands.postLink
     ], next)
   ), 
   callback
);
```

Kureev:
I. JUST. CANT. BELIVE!

grabbou:
Wait for your first contract and you will know that pain haha!

Kureev:
Released `1.4.0`

PR #20, opened 2016-01-20
Update deps

grabbou:


Kureev:
Thanks! Hope everything works fine :+1:

grabbou:
Tests green so there's no way it fails :D

PR #22, opened 2016-01-24
Support react-native 0.18+

Kureev:
## Story:
- Detect an installed version of the `react-native`
- Based on the installed version we're using a specific patch

This PR introduce two versions of the patch: for `react-native` < 0.18.0 and `react-native` > 0.18.0.

Should fix https://github.com/rnpm/rnpm/issues/77


Psykar:
Awesome, thank :)

grabbou:
Have you tested that? Is this an obligatory update? (All 0.18 projects have to use it?)

src/android/patches/0.17/index.js
```diff
@@ -0,0 +1,94 @@
...
+const MAIN_ACTIVITY_PACKAGE_PATTERN = `.addPackage(new MainReactPackage())`;
+
+const readFile = (file) =>
+  () => fs.readFileSync(file, 'utf8');
+
+const writeFile = (file, content) => content ?
+  fs.writeFileSync(file, content, 'utf8') :
+  (c) => fs.writeFileSync(file, c, 'utf8');
+
+const replace = (scope, pattern, patch) =>
```
grabbou: Any chances you extract common methods? These files seem to be super similar.

src/android/registerNativeModule.js
```diff
@@ -1,94 +1,14 @@
...
-  /**
-   * Check if module has been installed already
-   */
-  const isInstalled = compose(
-    (content) => ~content.indexOf(getMainActivityPatch()),
-    readFile(projectConfig.mainActivityPath)
+module.exports = function registerNativeModule(name, dependencyConfig, projectConfig) {
+  const rnpkg = require(
+    path.join(projectConfig.folder, 'node_modules', 'react-native', 'package.json')
   );
```
grabbou: Obviously this will be cached, but since we read package json anyway in `link` & `unlink` to get dependencies, you could just pass the version to the function w/o calling require. Should be easier to test.

Kureev:
Yes for both questions. Ill abstract a common patterns late

grabbou:
Also please add some unit tests and that should be all. Use mock-require and sinon.stub and based on package.json version passed to function, check what function was called.

Kureev:
Please, take a look, @grabbou :smiley: 
I have some plans to do some additional testing scenarios, but it shouldn't block shipping this feature.

As we agreed, I've tried to do my best separating a reusable parts of the code. Now it should be fine. Maybe, patching settings and build will go away (eventually) and the code will look even more clean. 

Anyway, if you have no strong objections about merging it in, I'd like to do it asap (don't want to block Android developers using it with a latest react-native).

src/android/getPrefix.js
```diff
@@ -0,0 +1,16 @@
...
+    path.join(config.folder, 'node_modules', 'react-native', 'package.json')
+  );
+
+  var prefix = 'patches/0.18';
+
+  if (semver.lt(rnpkg.version, '0.18.0')) {
+    prefix = 'patches/0.17';
+  }
+
+  return prefix;
```
grabbou: Can we inline that to const prefix = semver.... ? '' : '' ?

src/android/patches/0.18/makeMainActivityPatch.js
```diff
@@ -0,0 +1,26 @@
+const append = (scope, pattern, patch) =>
+  scope.replace(pattern, `${pattern}${patch}`);
+
+module.exports = function makeMainActivityPatch(config) {
+  const importPattern = 'import com.facebook.react.ReactActivity;';
+  const packagePattern = 'new MainReactPackage()';
+  /**
```
grabbou: line break? ;)

grabbou:
I know, i know. But still - I would like to keep the quality of the code even if we are about to do quick hot-fixing. It looks good to me. If I won't be around later today - feel free to merge & :ship: Anyway - I just added two questions to style that I personally feel we could easily improve and that's all.

src/android/getPrefix.js
Kureev: I'd prefer to keep it this way for the future versions. Ternary operator works fine in this case, but if in the next react-native version facebook will introduce something else, then we'll need to roll our back to the this one.

Kureev: Anyway, maybe in the future I'll come up with a better solution for this, but now it's just fine.

PR #25, opened 2016-01-29
Feature/android various stuff

grabbou:
A bunch of tweaks that we were talking about. I have that feeling you will rewrite most of the naming things I've done, but no worries - I had no clue about them, so tried to do my best ;)

What's done:

- [X] `getPrefix` now correctly reports patches prefix for release candidate versions. Previously (as of master), it would return old patches for 0.18-rc, 0.19-rc and other releases in the future.
- [X] I've added support for rn 0.18 in unlink - not sure if that's preferred way of discovering `mainActivityPatches`, but tried to extract them to common place so it's easier to keep them in sync
- [X] Moved out readFile/writeFile since we were using them in few places.
- [X] Get prefix accepts plain version instead of entire config object so it's easier to test (I've added more test cases and that helped me to find these edge cases described in first point)
- [ ] There is a PR in rnpm to pass rnVersion with a `projectConfig`, so then we can remove getReactNativeVersion entirely

Anyway, hope you'll find that useful, refactor it, merge & we ship next release! Good luck :muscle:


Kureev:
Can you please split it up for a small PRs?

Kureev:
Also, I'm not sure why we need to have a `addPackagePatch` in the separated file

src/android/registerNativeModule.js
```diff
@@ -1,16 +1,10 @@
-const fs = require('fs-extra');
+const fs = require('./fs');
```
Kureev: can't we use 
```js
const writeFile = require('./fs').writeFile;
const readFile = require('./fs').readFile;
```
and not to touch lines 7, 18, 20, 24, 26, 30, 32, 40?

grabbou: Probably better to clean up the diff, will do

grabbou:
> Can you please split it up for a small PRs?

Why? Is 70+ lines added too much to review? Only 3 significant updates that are somehow connected to each other ;)

> Also, I'm not sure why we need to have a addPackagePatch in the separated file

As I said, it's up to you. I just thought it's better since we rely on it in two places (unlink & mainActivityPatch), but I have no opinion on that.

Kureev:
:rainbow:! :ship: it!

grabbou:
Needs testing tomorrow #toolate

PR #28, opened 2016-02-06
Bug/wrong settings path

Kureev:
Currently we have a wrong `settings.gradle` path detection. 

We're using `path.relative(project.sourceDir, dependency.sourceDir)`, but `project.sourceDir` sometimes refers to `android/app` and sometimes to `android`. We don't reflect it in our patch code (yet).


grabbou:
But we already export settings.gradle path from config here -> https://github.com/rnpm/rnpm/blob/master/src/config/android/index.js#L42

Kureev:
Hm, I think I have no cases when settings.gradle is in the android/app. It means that the simplest solution would be just to use a `settingsGradlePath` and revert `isFlat` flag (doesn't make any sense?)

grabbou:
Yeah I mean we have been using that already! I think if settingsGradlePath might be wrong - it's rnpm where we should fix that since that plugin links correctly (?). I am fine with `isFlat` flag, might be useful at some point hehe :fire:

Kureev:
> I have project structure with app folder inside and I don't have settings.gradle in that folder (I have it in android folder directly). So that PR is going to break it. That's the structure created by react-native cli.

It wouldn't. By using `isFlat` I detect where it's located (see https://github.com/rnpm/rnpm-plugin-link/pull/28/files#diff-3485f024f7f244d8060cf04ba5ae6205R4)

grabbou:
```js
const projectRoot = projectConfig.isFlat
 +    ? path.join(projectConfig.sourceDir, '..')
 +    : projectConfig.sourceDir;
```
isFlat = true
sourceDir = 'home/android'
path = 'home/android/...'
settings.gradle is in home/android

Path will be wrong. I think since settings.gradle is always in `android` no matter structure, we are fine with sticking with how it worked.

Kureev:
> isFlat = true
sourceDir = 'home/android'
path = 'home/android/...'
settings.gradle is in home/android

Yes, but path to our module would be `../node_modules/<name>`, so it's correct

grabbou:
Well, if `path` is `home/android/../` - path to node_modules will be `./node_modules/<name>`. I think sticking with settings.gradle is better - at least it makes sense path.relative-wise.

Kureev:
Yeah, no doubts - I'm changing it to settingsGradlePath right now. Just saying that this one also works :)

Kureev:
Check it out :wink:

src/android/patches/makeSettingsPatch.js
```diff
@@ -2,7 +2,7 @@ const path = require('path');
 
 module.exports = function makeSettingsPatch(name, dependencyConfig, projectConfig) {
   const relative = path.relative(
-    path.dirname(projectConfig.settingsGradlePath),
+    projectConfig.settingsGradlePath,
```
grabbou: That has to be dirname, because path.relative given the path to file, will generate wrong relative path.

Kureev: Actually, it generates a correct path. You may double check it by running mocha

test/android/patches/makeSettingsPatch.spec.js
```diff
@@ -5,17 +5,19 @@ const expect = chai.expect;
 const makeSettingsPatch = require('../../../src/android/patches/makeSettingsPatch');
 
 const name = 'test';
-const projectConfig = {
-  sourceDir: '/home/project/android/app',
-  settingsGradlePath: '/home/project/android/settings.gradle',
-};
-const dependencyConfig = { sourceDir: `/home/project/node_modules/${name}/android` };
+const projectConfig = { settingsGradlePath: 'some/path/android' };
```
grabbou: SettingsGradlePath is `some/path/android/settings.gradle` - that's sent by rnpm link (as in previous unchanged version)

grabbou:
Since that PR is not actually changing anything, the question is do we still need it?

Kureev:
Before this PR we used `projectConfig.sourceDir`, after we're using `projectConfig.settingsGradlePath` so yes, we need it.

grabbou:
Are you sure?
screen shot 2016-02-06 at 23 21 11

src/android/patches/makeSettingsPatch.js
grabbou: As I already said, mocha tests are corrupted because you have changed the line "10" (settingsGradlePath pointing to a file) to line "8" where it points to a folder (which is wrong, because rnpm sends over a **path**, not **location**)

Issue #27, opened 2016-02-04
[Android] Not detecting that the plugin has already been linked


geof90:
When I try to link the same plugin twice, it shows this message which seems good:
```
rnpm info Linking react-native-code-push android dependency 
rnpm info Android module react-native-code-push is already linked 
rnpm info Linking react-native-code-push ios dependency 
rnpm info iOS module react-native-code-push is already linked
```

However when i look at the settings.gradle file in Android, it looks like it is not detecting that it is already linked and it is inserting the same line twice
eg.
```
include ':app'
include ':react-native-code-push'
project(':react-native-code-push').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-code-push/android/app')
include ':react-native-code-push' // <- this appears twice
project(':react-native-code-push').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-code-push/android/app') // <- this appears twice
```

geof90:
After more investigation, I see that this only happened to a project that was previously manually linked, not via RNPM.

So for example, this was my original gradle file
```gradle
include ':app', ':react-native-code-push' // note that i am including two projects in one line
project(':react-native-code-push').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-code-push/android/app')
```
after running `rnpm link`, it becomes this invalid gradle file:

```gradle
include ':app'
include ':react-native-code-push'
project(':react-native-code-push').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-code-push/android/app'), ':react-native-code-push'
project(':react-native-code-push').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-code-push/android/app')
```

It looks like it is simply doing a string insertion right after `include ':app'`.

grabbou:
cc: @Kureev probably related to the way you detect if app is installed? Maybe some regexp would be helpful here?

Kureev:
well, it's because of `, ':react-native-code-push'` at the end of the line.

lostintangent:
@Kureev Isn't is totally legitimate for the include statement to include a comma separated list of projects in addition to calling include N number of times? It would be awesome if the link command supported integration into existing projects that were using either syntax.

geof90:
After looking at the code, I realize that the linkage detection code does not look at the gradle file at all, but looks at the MainActivity. I have a PR up that could potentially fix this for most users, please take a look.

grabbou:
[...]

Issue #26, opened 2016-01-31
Make MainActivity detection more solid


grabbou:
It's not a requirement to use new MainActivity pattern starting from React 0.18 and some people might still keep using the old one (I don't know why, but it can happen).

Instead of checking for react version, let's check if MainActivity class extends "Activity" (that's the old version) or "ReactActivity" (that's the new version). Will give us best answer.

Kureev:
yeah, good proposal

geof90:
Actually, instead of looking for the import pattern, can we look for the MainActivity class definition pattern instead? So e.g.

https://github.com/rnpm/rnpm-plugin-link/compare/master...geof90:patch-2

This would work for both <0.17 and >=0.18. We could easily just use the same file.

Kureev:
But patches for main activity looks different for <0.17 and 0.18+ versions

grabbou:
We could probably make only different `packagePatches` instead of having entire mainActivity patches that way.

Kureev:
Yes, we can improve this by making a change scope more atomic. Instead of the whole file we can change only one part.

geof90:
Yup, using different package patches but using the same import patch was what I was trying to suggest. My code sample was just to communicate the idea, its not working code :p

grabbou:
@geof90 it's been a while since we discussed it, but would you still be interested in taking that? I am not that much into these patches @Kureev done but I quite like the idea of making it slicker!

geof90:
sure, I'll take a look over the weekend!

grabbou:
After great @Kureev work on patches and the general idea to deprecate them at some point (0.17 is really old and probably will be no longer used in like a year), we can close it and revisit if we will be about to add yet another `patch`.

PR #29, opened 2016-02-06
More foolproof way of checking plugin installation

geof90:
Fixes https://github.com/rnpm/rnpm-plugin-link/issues/27.

The original way of checking for the presence of `new PackageClassName()` inside the MainActivity could fail in the following scenarios:

1. If the PackageClassName object requires extra parameters in construction, e.g `new PackageClassName(params...)`;
2. If the `PackageClassName` object is a private instance and not directly constructed, e.g. https://github.com/Microsoft/react-native-code-push/blob/master/Examples/CodePushDemoApp/android/app/src/main/java/com/microsoft/codepushdemoapp/MainActivity.java#L54

In general, looking into Java code to detect linkage is not foolproof because code is relatively non static and is always changed over the lifetime of a project.

This change proposes looking into the `build.gradle` file to detect linkage instead.

### DIFF
--- a/src/android/registerNativeModule.js
+++ b/src/android/registerNativeModule.js
@@ -37,8 +37,8 @@ module.exports = function registerNativeAndroidModule(name, dependencyConfig, pr
    * Check if module has been installed already
    */
   const isInstalled = compose(
-    (content) => ~content.indexOf(dependencyConfig.packageInstance),
-    readFile(projectConfig.mainActivityPath)
+    (content) => ~content.indexOf(`:"${name}"`) || ~content.indexOf(`':${name}'`),
+    readFile(projectConfig.buildGradlePath)
   );
 
   if (!isInstalled(name)) {


geof90:
cc @cbrevik

grabbou:
Looks really good to me ;) Let's wait for @Kureev.

PS. Once you are on that, would you mind checking https://github.com/rnpm/rnpm-plugin-link/issues/26#issue-130116117 as well?

geof90:
@grabbou do you want me to include a fix for that in this same PR?

grabbou:
I don't really mind, but as long as I know @Kureev, he'd prefer having it separate :fire:

Kureev:
I'll check it tomorrow. Thanks for the PR :heart:

geof90:
@Kureev No probs! We really excited to use your great work once the right enhancements are in place.

Kureev:
LGTM :rainbow:

cbrevik:
Been away on a trip, didn't see this till now. Even more clever solution looking in build.gradle, nice! :+1:

cbrevik:
Though, it is still vulnerable to alternative naming schemes. People can define whatever name for the library project they want in `android/settings.gradle`, which would invalidate the `android/app/build.gradle` check.

E.g. `settings.gradle`:
```gradle
include ':GoogleAnalyticsBridge'
project(':GoogleAnalyticsBridge').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-google-analytics-bridge/android')
```
`build.gradle`:
```gradle
dependencies {
...
    compile project(':GoogleAnalyticsBridge')
}
```

And checking for `":react-native-google-analytics-bridge"`. Granted, this might not be a big problem.

PR #36, opened 2016-02-29
Make patch work on windows

terebentina:
path.sep on windows is `\`, breaking the path in settings.gradle. The fix is to either duplicate it as in `..\\some\\path` or replace it with `/` which still works on windows.

fixes rnpm/rnpm-plugin-link#36

### DIFF
--- a/src/android/patches/makeSettingsPatch.js
+++ b/src/android/patches/makeSettingsPatch.js
@@ -4,7 +4,7 @@ module.exports = function makeSettingsPatch(name, dependencyConfig, projectConfi
   const relative = path.relative(
     path.dirname(projectConfig.settingsGradlePath),
     dependencyConfig.sourceDir
-  );
+  ).replace(/\\/g, '/');
 
   /**
    * Replace pattern by patch in the passed content


Kureev:
Hi, sorry for the late response, but why `path.relative` doesn't work for you? It should work for windows machines also: https://nodejs.org/api/path.html#path_path_relative_from_to

terebentina:
`path.relative` works just fine, that's not the problem. If I `console.log(relative)` as the first line in `applySettingsPatch`, it correctly shows the path with double backslashes. However, when you stringify it again inside the template literal the double backslashes are transformed into single backslashes and this breaks everything.
Here, try this 1 line in a devtools console:
```javascript
var asd = '\\hi\\there'; console.log(`${asd}`)
```
You will see that the output has only single backslashes (`\hi\there`).

Kureev:
You mean that by using `` we accidentally break a version for windows? Then it sounds like we need to replace `` by '' instead of replacing \\ by /, right?

terebentina:
Could you edit your comment and surround \ by backticks so that the comment is readable?

Kureev:
Fixed

terebentina:
I think using plain old string concatenation would work, indeed. Note that my PR replaces `\` with `/` not `\\` with `/`. That's the whole point - by the time of the replace, there's no `\\` anymore.

Kureev:
> by the time of the replace, there's no `\\` anymore

I believe that's the underlying issue we should fix.

terebentina:
Meh, `'str' + 'str'` doesn't work either - it still transforms `\\` into `\`.
I still think my fix is the quickest/least intrusive. `/` works on windows just fine so it's a cross-platform solution.
The alternative to this, as I mentioned in my first comment, would be to replace `\` with `\\`. So my replace in this PR would become `.replace(/\\/g, '\\\\');` so that those 4 backslashes become 2 in `applySettingsPatch`. 
Since 1 slash looks better than 4 backslashes, I opted for it.

terebentina:
Without my PR:
```javascript
include ':react-native-material-kit'
project(':react-native-material-kit').projectDir = new File(rootProject.projectDir, '..\node_modules\react-native-material-kit\android')
include ':react-native-select-contact-android'
project(':react-native-select-contact-android').projectDir = new File(rootProject.projectDir, '..\node_modules\react-native-select-contact-android\android')
include ':react-native-vector-icons'
project(':react-native-vector-icons').projectDir = new File(rootProject.projectDir, '..\node_modules\react-native-vector-icons\android')
```

With my PR:
```javascript
include ':react-native-material-kit'
project(':react-native-material-kit').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-material-kit/android')
include ':react-native-select-contact-android'
project(':react-native-select-contact-android').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-select-contact-android/android')
include ':react-native-vector-icons'
project(':react-native-vector-icons').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-vector-icons/android')
```

With the 4 backslashes (`.replace(/\\/g, '\\\\');`)
```javascript
include ':react-native-material-kit'
project(':react-native-material-kit').projectDir = new File(rootProject.projectDir, '..\\node_modules\\react-native-material-kit\\android')
include ':react-native-select-contact-android'
project(':react-native-select-contact-android').projectDir = new File(rootProject.projectDir, '..\\node_modules\\react-native-select-contact-android\\android')
include ':react-native-vector-icons'
project(':react-native-vector-icons').projectDir = new File(rootProject.projectDir, '..\\node_modules\\react-native-vector-icons\\android')
```

Kureev:
Ok, if this works for you (and hopefully to other Windows users), then I'm fine with merging it in. But to avoid some confusion, I'd like to ask you to make it a platform-specific code:

Instead of
```js
const relative = path.relative(
  path.dirname(projectConfig.settingsGradlePath),
  dependencyConfig.sourceDir
).replace(/\\/g, '/');
```
it would be nice to have
```js
let relative = path.relative(
  path.dirname(projectConfig.settingsGradlePath),
  dependencyConfig.sourceDir
);

if (process.platform === 'win32') {
  // fix broken paths for Windows users
  relative = relative.replace(/\\/g, '/');
}
```
or something similar.

grabbou:
Yeah, plus instead of `// fix broken paths for Windows users` let's add a proper description `WHY` we do it (or link to this PR, both are fine)

Kureev:
@terebentina would you mind to update a PR?

terebentina:
Oh, right, will do tomorrow

grabbou:
Going to close this one since we've merged #39 - thanks for your contribution to make it better on Windows!

PR #40, opened 2016-03-19
Bump glob

grabbou:
No changes needed.


Kureev:
I'd say we need to downgrade it (due to compatibility with existing react-native glob version):
image

grabbou:
I can see there's already 7.0.0 installed somewhere, so it's likely not going to improve anything?

Kureev:
hm, mb you're right. Go for it!

PR #46, opened 2016-03-23
Android: isInstalled fails for package entries with double quotes

flipchart:
The `.indexOf` pattern was wrong: the colon was outside the quotes

### DIFF
--- a/src/android/registerNativeModule.js
+++ b/src/android/registerNativeModule.js
@@ -37,7 +37,7 @@ module.exports = function registerNativeAndroidModule(name, dependencyConfig, pr
    * Check if module has been installed already
    */
   const isInstalled = compose(
-    (content) => ~content.indexOf(`:"${name}"`) || ~content.indexOf(`':${name}'`),
+    (content) => ~content.indexOf(`":${name}"`) || ~content.indexOf(`':${name}'`),
     readFile(projectConfig.buildGradlePath)
   );
 


grabbou:
Hey!

What about the second check that does have quotes?
` ~content.indexOf(`':${name}'`

flipchart:
Before:

    (content) => ~content.indexOf(`:"${name}"`) || ~content.indexOf(`':${name}'`),
                                   ^  outside quotes

After:

    (content) => ~content.indexOf(`":${name}"`) || ~content.indexOf(`':${name}'`),
                                    ^ inside quotes

The second check is fine because the colon is inside the (single) quotes. The first check is looking for double quotes.

grabbou:
Ah, true, entirely missed that!

@Kureev do we need quotes in the second check? Can't it simply be `:${name}` ?

Kureev:
Actually, we always need to check for this pattern:
https://github.com/rnpm/rnpm-plugin-link/blob/master/src/android/patches/makeBuildPatch.js#L9

Kureev:
Simply put, we need to check for `:${name}`

grabbou:
Anyone wants to make a PR?

Kureev:
I'm planning to round off my side activities before the weekend and resume working on rnpm. 
That would be a first thingy I'll tackle if nobody submit a PR before.

grabbou:
Closing in favour of #49 that slightly unifies these checks since `"` and `'` are irrelevant! Thanks for bringing this up to the table - it's been a valuable hint! I hope the fact we are merging another PR does not discourage you from further contributing to this project! The more Android heros here the better for all of us!

PR #48, opened 2016-03-24
Fix error when the header search path is only a string, not an array

cmcewen:
Should fix https://github.com/rnpm/rnpm/issues/117


grabbou:
Thanks! One thing I am wondering - can we maybe use `Array.isArray` instead?

Kureev:
@cmcewen would you mind to update a PR?

PR #49, opened 2016-03-24
Follow up to #46

grabbou:
Just submitting a PR - I think that change should not hurt anyone. Maybe we could use regex to find a `: ("|') ${name} ("|')` (stuff in brackets is optional)

### DIFF
--- a/src/android/registerNativeModule.js
+++ b/src/android/registerNativeModule.js
@@ -37,7 +37,7 @@ module.exports = function registerNativeAndroidModule(name, dependencyConfig, pr
    * Check if module has been installed already
    */
   const isInstalled = compose(
-    (content) => ~content.indexOf(`:"${name}"`) || ~content.indexOf(`:${name}`),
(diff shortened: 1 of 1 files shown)


Kureev:
We can, but we don't need to. Our code always links it using one pattern: `:react-native-plugin-name` (without " or ') if you want a more generic pattern, then remove `:` in general. If you have a plugin name in the file, it's installed.

Kureev:
I wanna say that `:"something"` is never a correct pattern.

grabbou:
Ah, cool. You are right with the double quote - haven't checked the gradle when suggesting that :D

Kureev:
Then seems I'm doing it wrong now :smiley:

grabbou:
Yeah I just checked and it's:
```
    compile project(":react-native-fs")
    compile project(":react-native-google-analytics-bridge")
    compile project(":react-native-linear-gradient")
    compile project(":RNMail")
```
in one of my projects.

PR #50, opened 2016-03-28
Fixes #21

### LINKED ISSUE #21: Improve `unlinkAssets`
When linking assets, we build an array of UNIQUE files, so that if both packages use the same font, we dedupe and copy just one of them.

When unlinking - with the current version, we will remove the fonts just because the package was using them, without checking if there are any other packages that are using it.

Solution:
- getAllDependencies
- getAssets (unique)
- difference of currentDependency.assets & ^
- unlink the stuff from the above function

I will send it over later this week (over the weekend). Does not seem like a deal breaker as that situation is rare to happen in general.

This will require getAssets & getDependencies to be striped out from link and moved to some reusable methods (so we do not duplicate them in `unlink`) making them easily testable as well.

grabbou:


grabbou:
@appden the only case we might need to work around here is to how to handle removing these things in `unlink`, to not produce conflicts with other dependencies.

The most obvious way would've been to get all dependencies config and just check if they have these things defined - if yes - do not delete. Same for `add`. Just wondering if there's a better way (we can implement `link` now and think about it later)

appden:
@grabbou thanks for scaffolding this out! I'll take a look at implementing it later today. You're right that in order to remove things during `unlink`, we need to know what all the other dependencies are. Since there's no authoritative place where `rnpm` stores that information, it probably would be best to look at the linked static libraries, and deduce from there based on the `node_modules/<name>` in the path of its containing Xcode project.

grabbou:
Yeah - you are right. Keep in mind that also there's #21 - I am actually thinking I might add it to this branch in next commit just to make the `get all dependencies` part already resolved.

grabbou:
@appden I've decided to rename this to be `fixes #21` so we can merge it in faster (and not delay / interrupt your work). It also has now a clear place for `allDependencies` so you can do the flatMap just as I did for assets.

@Kureev this one is ready for review and treat it as a fix to #21

src/getDependencyConfig.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Given an array of dependencies - it returns their RNPM config
+ * if they were valid.
+ */
+module.exports = function getDependencyConfig(config, deps) {
```
grabbou: @Kureev not a huge fan of how this function is named, but couldn't think of anything better. Feel free to update it!

src/ios/removeSharedLibraries.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function removeSharedLibraries(project, libraries) {
+
+};
```
grabbou: These two functions are just dummy placeholders that we can merge to master since there's an ongoing work to implement them now. Since this PR was meant to be a base starter for them, I don't want to remove them to avoid conflicts later.

src/unlink.js
```diff
@@ -63,7 +69,7 @@ module.exports = function unlink(config, args) {
     }
   }
 
-  const assets = dependency.assets;
+  const assets = diff(dependency.assets, flatMap(allDependencies, d => d.assets));
```
grabbou: A good task for tomorrow I'll have to do is to write unit tests for `unlink` part (just as we did for `link`).

src/flatMap.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function flatMap(arr, mapper) {
+  return arr.reduce(
+    (acc, item) => acc.concat(mapper(item)),
+    []
+  );
+};
```
Kureev: we already use lodash, why?

src/unlink.js
```diff
@@ -2,10 +2,14 @@ const path = require('path');
 const log = require('npmlog');
 
 const isEmpty = require('./isEmpty');
+const getProjectDependencies = require('./getProjectDependencies');
 const unregisterDependencyAndroid = require('./android/unregisterNativeModule');
 const unregisterDependencyIOS = require('./ios/unregisterNativeModule');
 const unlinkAssetsAndroid = require('./android/unlinkAssets');
 const unlinkAssetsIOS = require('./ios/unlinkAssets');
+const getDependencyConfig = require('./getDependencyConfig');
+const diff = require('./diff');
```
Kureev: btw, I think diff can be also replaced by lodash function

Kureev:
I think we're good to go

src/flatMap.js
grabbou: In that PR we didn't :P

PR #51, opened 2016-03-29
Replace self-written helpers by lodash functions

Kureev:
subj

(as we discussed in messenger)


src/unlink.js
```diff
@@ -1,15 +1,13 @@
...
-const isEmpty = require('./isEmpty');
 const getProjectDependencies = require('./getProjectDependencies');
 const unregisterDependencyAndroid = require('./android/unregisterNativeModule');
 const unregisterDependencyIOS = require('./ios/unregisterNativeModule');
 const unlinkAssetsAndroid = require('./android/unlinkAssets');
 const unlinkAssetsIOS = require('./ios/unlinkAssets');
 const getDependencyConfig = require('./getDependencyConfig');
-const diff = require('./diff');
-const flatMap = require('./flatMap');
+const _ = require('lodash');
```
grabbou: Isn't it better to require each function separately? Just to make it clearer what do we use (that's mainly a personal preference)

grabbou:
Okay, cool. That looks good. I'll double check in the morning and merge by then! Thanks!

PR #53, opened 2016-03-30
Handle nested groups

grabbou:
What this PR does:

- You can specify nested `libraryFolder`, see my comment below,
- When you have multiple groups with the same name, we no longer pick the wrong one, see: https://github.com/rnpm/rnpm/issues/111#issuecomment-202984773

Technical explanation:
Problem is - `pbxGroupByName` will return first matching group by name (regardless of nesting level). We loop recursively over all groups with our new function and also support paths, e.g. `NestedGroup/Libraries`, so if you would prefer to use libraries group that's not top-level, you can easily override that.

That PR also implements the above for `createGroup`.

### DIFF
--- a/src/ios/createGroup.js
+++ b/src/ios/createGroup.js
@@ -3,9 +3,8 @@ const getGroup = require('./getGroup');
 const hasGroup = (pbxGroup, name) => pbxGroup.children.find(group => group.comment === name);
 
 /**
- * Given project and name of the group, creates a new top-level group,
- * that is then - added to `mainGroup` so that it's visible as a top-level
- * folder in your Xcode sidebar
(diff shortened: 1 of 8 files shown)


grabbou:
```
rnpm-link info Linking react-native-blur ios dependency
rnpm-link WARN ERRGROUP Group Mike/Test/Libraries does not exist in your XCode project. We have created it automatically for you.
rnpm-link info iOS module react-native-blur has been successfully linked
```

```json
  "rnpm": {
    "ios": {
      "libraryFolder": "Mike/Test/Libraries"
    }
  }
```

screen shot 2016-03-30 at 09 33 19

PR #55, opened 2016-03-31
getPlistPath return value fix

rahimsiz:
This PR fixes getPlistPath return value, at current version it returns null or undefined


grabbou:
Thank you!!

PR #56, opened 2016-04-01
Chore/dynamic require error handling

Kureev:
- [x] Prevent error swallowing from the dynamically required modules

**Side-track fix:** correct repo details in the `package.json`. 
Feel free to ask me to exclude it form this PR (out of scope).

### DIFF
--- a/src/android/registerNativeModule.js
+++ b/src/android/registerNativeModule.js
@@ -17,7 +17,8 @@ module.exports = function registerNativeAndroidModule(name, dependencyConfig, pr
     const makeBuildPatch = require(`./patches/makeBuildPatch`);
     const makeMainActivityPatch = require(`./${prefix}/makeMainActivityPatch`);
   } catch (e) {
-    log.error(e + '.\nPlease file an issue here: ' + pkg.bugs.url);
+    log.error(e + '.\nIt seems somethins went wrong while patching ' +
+      'Android modules.\nPlease file an issue here: ' + pkg.bugs.url);
   }
 
   const performSettingsGradlePatch = applyPatch(
--- a/src/android/registerNativeModule.js
+++ b/src/android/registerNativeModule.js
@@ -17,7 +17,7 @@ module.exports = function registerNativeAndroidModule(name, dependencyConfig, pr
     const makeBuildPatch = require(`./patches/makeBuildPatch`);
     const makeMainActivityPatch = require(`./${prefix}/makeMainActivityPatch`);
   } catch (e) {
-    log.error(e + '.\nIt seems somethins went wrong while patching ' +
+    log.error(e + '.\nIt seems something went wrong while patching ' +
       'Android modules.\nPlease file an issue here: ' + pkg.bugs.url);
   }
 
--- a/src/android/registerNativeModule.js
+++ b/src/android/registerNativeModule.js
@@ -1,7 +1,9 @@
+const log = require('npmlog');
+const compose = require('lodash').flowRight;
 const readFile = require('./fs').readFile;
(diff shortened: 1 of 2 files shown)


src/android/registerNativeModule.js
```diff
@@ -9,9 +11,15 @@ const applyPatch = (filePath, patch) =>
...
-  const makeMainActivityPatch = require(`./${prefix}/makeMainActivityPatch`);
+
+  try {
+    const makeSettingsPatch = require(`./patches/makeSettingsPatch`);
+    const makeBuildPatch = require(`./patches/makeBuildPatch`);
+    const makeMainActivityPatch = require(`./${prefix}/makeMainActivityPatch`);
+  } catch (e) {
+    log.error(e + '.\nIt seems somethins went wrong while patching ' +
+      'Android modules.\nPlease file an issue here: ' + pkg.bugs.url);
+  }
```
grabbou: Are we ok with logging from registerNativeModule functions as well? :D Isn't it better to reject from this function and log in a place where we do `dependencies.map(register).then(...)` ? :D

grabbou: BTW - really good idea!

Kureev: Unfortunately, the reason why I made it - is because you can't propagate anything from the dynamically required module up. Otherwise it wouldn't be even needed.

grabbou:
So the error is coming from the fact there's no such folder with given prefix or that file throws under some circumstances?

Kureev:
Nope, the error comes from the developer issue. I mean if we have a typo or variable doesn't exist.

Kureev:
Can we merge it in?

grabbou:
It's still unclear to me why we have a try catch there from plain require. On master we now run these functions inside a promise with a global catch at the end, so shouldn't that just work? m/rnpm-plugin-link/blob/master/src/link.js#L125-L131

Is there any issue or a reproduction I can try to see what's the problem that fixes?

Kureev:
If you compose a require path in the application lifecycle, node will swallow any errors that this require will throw.

Example:

a.js:
```js
some shit there
```

b.js:
```js
function dynamicRequire(filename) {
  return require('./' + filename);
}

dynamicRequire('a'); // no errors
```

Kureev:
> Is there any issue or a reproduction I can try to see what's the problem that fixes?

When something in dynamic require fails (like `./${prefix}/makeMainActivityPatch`), then it becomes really hard to debug.

grabbou:
Ok, thanks, will throw an error and see what happens 🎱

Kureev:
Did you managed to reproduce it?

Kureev:
I think I can probably come up with a better structural solution than this one. We can close it for now.

grabbou:
I put `throw new Error('Test')` in one of the patches then run `link` and I got proper stack trace. So I can't reproduce it.
```
/Users/grabbou/Repositories/rnpm/rnpm-plugin-link/src/android/patches/0.20/makePackagePatch.js:5
  throw new Error('Test');
  ^

Error: Test
    at makePackagePatch (/Users/grabbou/Repositories/rnpm/rnpm-plugin-link/src/android/patches/0.20/makePackagePatch.js:5:9)
    at unregisterNativeAndroidModule (/Users/grabbou/Repositories/rnpm/rnpm-plugin-link/src/android/unregisterNativeModule.js:40:5)
    at Object.unlink [as func] (/Users/grabbou/Repositories/rnpm/rnpm-plugin-link/src/unlink.js:46:30)
    at Command.runAction (/Users/grabbou/Repositories/rnpm/rnpm/bin/cli:21:15)
```

PR #57, opened 2016-04-01
Feature/param polling system

Kureev:
Ask user for the params that he can specify in the config:

```json
"rnpm": {
  "android": {
    "packageInstance": "new SomeLibName(this, ${first}, ${last})"
  },
  "params": [{
    "type": "input",
    "name": "first",
    "message": "What's your first name"
  }, {
    "type": "input",
    "name": "last",
    "message": "What's your last name"
  }]
}
```

if no params has been specified, then inquirer part would be omitted and `${}` variables would be translated to "undefined".

### DIFF
--- a/src/android/patches/0.20/makeMainActivityPatch.js
+++ b/src/android/patches/0.20/makeMainActivityPatch.js
@@ -3,12 +3,12 @@ const addPackagePatch = require('./addPackagePatch');
 const append = (scope, pattern, patch) =>
   scope.replace(pattern, `${pattern}${patch}`);
 
-module.exports = function makeMainActivityPatch(config) {
+module.exports = function makeMainActivityPatch(androidConfig, params) {
   const importPattern = 'import com.facebook.react.ReactActivity;';
   const packagePattern = 'new MainReactPackage()';
 
-  const packageInstance = config.packageInstance
-    .replace(/\$\{(\w+)\}/g, (pattern, paramName) => config.params[paramName]);
+  const packageInstance = androidConfig.packageInstance
+    .replace(/\$\{(\w+)\}/g, (pattern, paramName) => `"${params[paramName]}"`);
 
   /**
    * Make a MainActivity.java program patcher
@@ -20,7 +20,7 @@ module.exports = function makeMainActivityPatch(config) {
     const patched = append(
       content,
       importPattern,
-      '\n' + config.packageImportPath
+      '\n' + androidConfig.packageImportPath
     );
 
     return append(
--- a/src/android/patches/0.20/makeMainActivityPatch.js
+++ b/src/android/patches/0.20/makeMainActivityPatch.js
@@ -7,6 +7,9 @@ module.exports = function makeMainActivityPatch(config) {
   const importPattern = 'import com.facebook.react.ReactActivity;';
   const packagePattern = 'new MainReactPackage()';
 
+  const packageInstance = config.packageInstance
+    .replace(/\$\{(\w+)\}/g, (pattern, paramName) => config.params[paramName]);
+
   /**
    * Make a MainActivity.java program patcher
    * @param  {String}   importPath Import path, e.g. com.oblador.vectoricons.VectorIconsPackage;
@@ -23,7 +26,7 @@ module.exports = function makeMainActivityPatch(config) {
     return append(
       patched,
       packagePattern,
-      addPackagePatch(config)
+      addPackagePatch({packageInstance})
     );
   };
 };
--- a/src/android/patches/0.20/makeMainActivityPatch.js
+++ b/src/android/patches/0.20/makeMainActivityPatch.js
@@ -7,8 +7,10 @@ module.exports = function makeMainActivityPatch(androidConfig, params) {
   const importPattern = 'import com.facebook.react.ReactActivity;';
   const packagePattern = 'new MainReactPackage()';
 
-  const packageInstance = androidConfig.packageInstance
-    .replace(/\$\{(\w+)\}/g, (pattern, paramName) => `"${params[paramName]}"`);
+  const packageInstance = androidConfig.packageInstance.replace(
+    /\$\{(\w+)\}/g,
+    (pattern, paramName) => params[paramName] ? `"${params[paramName]}"` : null
+  );
 
   /**
    * Make a MainActivity.java program patcher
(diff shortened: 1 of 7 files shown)


Kureev:
I'll fix tests and update this PR. In the meanwhile you can take a look.

Kureev:
Ready for review

Kureev:
@grabbou ?

grabbou:
It's weekend, I'll be checking that out on Monday. I took a look briefly earlier today and it looks awesome, great work. I have a few ideas going forward so I can add in these things to iOS as well. I will write them down tomorrow :)

src/link.js
```diff
@@ -19,34 +20,46 @@ const promisify = (func) => () => new Promise((resolve, reject) =>
   func((err, res) => err ? reject(err) : resolve(res))
 );
 
+function promiseWaterfall(tasks) {
+  return tasks.reduce(
+    (prevTaskPromise, task) => prevTaskPromise.then(task),
+    new Promise((r, e) => r())
+  );
+}
```
grabbou: Can we move that to sep. file & test? :)

src/link.js
```diff
@@ -19,34 +20,46 @@ const promisify = (func) => () => new Promise((resolve, reject) =>
...
-      dependency.name,
-      dependency.config.android,
-      project.android
-    );
-
-    if (didLinkAndroid) {
-      log.info(`Android module ${dependency.name} has been successfully linked`);
-    } else {
-      log.info(`Android module ${dependency.name} is already linked`);
+  return new Promise((resolve, reject) => {
```
grabbou: Is it really necessary to wrap it in a promise in this PR? Also - since it's always resolved successfully, I think we can just leave it as is (and add that when we really need async behaviour here)

src/link.js
```diff
@@ -88,17 +100,23 @@ module.exports = function link(config, args) {
...
-    project.assets
-  ));
-
-  return tasks.then(() => linkAssets(project, assets));
+  return Promise.all(
+    dependencies.map(dependency => promiseWaterfall([
+      () => pollParams(dependency.config.params),
+      (params) => new Promise((res, rej) => {
+        dependency.config.params = params;
+        res();
```
grabbou: Same here, I don't think we need a promise here. we could just do the assignment, e.g. ` => void ( dependency.config.params = params)`

src/link.js
```diff
@@ -88,17 +100,23 @@ module.exports = function link(config, args) {
...
+      }),
+      () => promisify(dependency.config.commands.postlink || commandStub),
+      () => linkDependency(project, dependency),
+      () => promisify(dependency.config.commands.prelink || commandStub),
+    ]))
+  ).then(() => {
+    const assets = dedupeAssets(dependencies.reduce(
```
grabbou: Can we maybe make assets above line `103` and make this just `then(() => linkAssets(...))` ?

src/android/patches/0.20/makeMainActivityPatch.js
```diff
@@ -3,10 +3,13 @@ const addPackagePatch = require('./addPackagePatch');
 const append = (scope, pattern, patch) =>
   scope.replace(pattern, `${pattern}${patch}`);
 
-module.exports = function makeMainActivityPatch(config) {
+module.exports = function makeMainActivityPatch(androidConfig, params) {
   const importPattern = 'import com.facebook.react.ReactActivity;';
   const packagePattern = 'new MainReactPackage()';
 
+  const packageInstance = androidConfig.packageInstance
+    .replace(/\$\{(\w+)\}/g, (pattern, paramName) => `"${params[paramName]}"`);
```
grabbou: Can we move this to `src/<name_it.js>` and unit-test so I can reuse that? :D What's the behaviour when no value is given for a param? will it e.g. be `undefined` ? I think it would make sense to use a `null` since that's what works in Java.

src/link.js
```diff
@@ -19,34 +20,46 @@ const promisify = (func) => () => new Promise((resolve, reject) =>
   func((err, res) => err ? reject(err) : resolve(res))
 );
 
+function promiseWaterfall(tasks) {
+  return tasks.reduce(
+    (prevTaskPromise, task) => prevTaskPromise.then(task),
+    new Promise((r, e) => r())
```
grabbou: This could be `Promise.resolve()`

src/pollParams.js
```diff
@@ -0,0 +1,13 @@
+var inquirer = require('inquirer');
+
+module.exports = (questions) => new Promise((resolve, reject) => {
+  if (!questions) {
+    return resolve({});
+  }
+
+  try {
+    inquirer.prompt(questions, resolve);
```
grabbou: PS. When that can throw?

src/pollParams.js
```diff
@@ -0,0 +1,13 @@
...
+  } catch (e) {
+    reject(e);
```
grabbou: I don't think we need a try/catch in a promise, if anything throws here, it will be passed to reject resolving function by default.

grabbou:
All in all looks great! Thanks for coming up with that idea! Left few comments and we are ready to go!

src/android/patches/0.20/makeMainActivityPatch.js
Kureev: I don't think we need to allocate a standalone file just for regexp. Besides that, even if we'll do it, there are some differences (as you said), like nil vs null. 

We can test the whole functionality in one place without creating a tons of 1-10-line files, so I don't see a point there also.

grabbou: Ok I am fine with that.

What about replacing undefined with `null`?

Kureev: Yeah, make sense. I'll add it today

Kureev:
Updated 😉

grabbou:
Thanks, I'll take a look in the evening! #exciting

lostintangent:
This is awesome! I would update the CodePush plugin ASAP when this is published to NPM, since it would make the Android setup even easier. You guys are killing it :)

grabbou:
Thanks @lostintangent :) Credits to Alexey for coming up with this!

@Kureev are you happy with how it works? Code-wise all seems good! Would be cool to ship it along the other fix here ( #58 )

Kureev:
Nice, merge it in when you're ready

grabbou:
Related: https://github.com/rnpm/rnpm/commit/2cc3ea7d7e1839163cdd5b7c7fd8585ed3b1d3a4 @Kureev

lostintangent:
FYI, I just submitted a PR to CodePush to use the new params feature. It makes setting up Android way nicer! Now if we can just add `Info.plist` entries using a param value as well that would make iOS even better :)

PR #58, opened 2016-04-05
Fixes #106 #134

grabbou:
Explanation tomorrow - still investigating.

Todo: there are still <s>two</s> (only one, 2nd one fixed) extra places we have to clean up

Now it all works well - just need to make sure diff is clean.

### DIFF
--- a/src/ios/removeFileFromProject.js
+++ b/src/ios/removeFileFromProject.js
@@ -1,16 +1,5 @@
 const PbxFile = require('xcode/lib/pbxFile');
-
-const removeContainerProxy = (project, file) => {
-  const section = project.hash.project.objects.PBXContainerItemProxy;
-
-  for (var key of Object.keys(section)) {
-    if (section[key].containerPortal_comment === file.basename) {
-      delete section[key];
-    }
-  }
-
-  return;
-};
+const removeFromPbxItemContainerProxySection = require('./removeFromPbxItemContainerProxySection');
 
 /**
  * Given xcodeproj and filePath, it creates new file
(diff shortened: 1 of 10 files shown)


grabbou:
PS. Last commits are not showing here for me - might be Github issue, will trigger another commit later.

src/ios/removeFromStaticLibraries.js
```diff
@@ -14,6 +15,8 @@ module.exports = function removeFromStaticLibraries(project, path, opts) {
   project.removeFromPbxBuildFileSection(file);
   project.removeFromPbxFrameworksBuildPhase(file);
   project.removeFromLibrarySearchPaths(file);
+  project.removeFromProductsPbxGroup(file);  // @todo this does not work but it should
```
Kureev: Fix this, cover by tests (optionally, if that's a hotfix) and we're ready to go 👍

grabbou: It's something I have to understand a bit deeper since technically speaking we have ~10 `Products` groups so this method is for a slightly different use-case (however it's named in a similar way)

src/ios/removeFileFromProject.js
```diff
@@ -13,7 +14,13 @@ const removeFromProjectReferences = require('./removeFromProjectReferences');
  */
 module.exports = function removeFileFromProject(project, filePath) {
   const file = project.removeFromPbxFileReferenceSection(new PbxFile(filePath));
+  const projectRef = removeFromProjectReferences(project, file);
+
+  if (projectRef) {
+    removeProductGroup(project, projectRef.ProductGroup);
```
grabbou: So @Kureev for each projectReference there's a productGroup created. Added it here, so I can just remove by ProductGroup uuid :)

grabbou:
Todo: add internal iterator

grabbou:
@Kureev I think it's ready, need to take a look at the final code once again but I am happy with how it works.

grabbou:
@Kureev ping me on FB when that's merged so I'll send a new release :) Today's a happy evening for our US users!

PR #59, opened 2016-04-06
Fix error when RN version is x.x.x-rc1

geof90:
Currently:
```javascript
let rnVersion = '0.24.0-rc1';
const version = rnVersion.replace('-rc', ''); // becomes '0.24.01'
...
     if (semver.lt(version, item + '.0') && nextVersion) // throws exception since version is not valid semver
```

Fixed this, and also updated the unit tests.


lostintangent:
Just hit this issue trying to use RN `0.24.0-rc1` :)

grabbou:
Ah, so awesome! Thank you, merging & including in the todays release with other fixes!

PR #62, opened 2016-04-08
For some reason it no longer returns value

grabbou:
… which always prints `is already linked` :(


grabbou:
Broken by this commit https://github.com/rnpm/rnpm-plugin-link/commit/801ae2fd0d436212870729207846835e0542a1c6#diff-e76d7672392aed2b927c3dedf1490560L96

PR #63, opened 2016-04-08
Improve params UX and fix broken unlink on Android

### LINKED ISSUE #61: Lazy poll for params
Currently, if you have 10 dependencies and each of them has params, rnpm will ask for them once, before even checking if they are already linked.

That means I'll have to answer all these questions every-time I run rnpm link.

We should poll for params inside `registerNativeModule` imo only for that specific dependency.

grabbou:
Lazily pools for params.

Fixes #61 #66 

What it fixes:
- do not poll for params if project is already linked
- do not poll for params for Android project if I don't have `android` in my project
- not working isInstalled check on Android (unlink) - replaced with extracted `isInstalled` function
- unit tests `isInstalled` Android & changes ~content.indexOf to be `>= 0` so that it always returns `boolean`, not a number (more explicit)

This PR:
- extracts two separate `linkDependency<Platform>` to separate functions that `can return a promise` which allows us to poll for params from inside them and also do some other async actions
- checks if dependency is installed before running registeryDependency - we no longer print `linking dependency... dependency is already linked` since we can check that upfront.
- flattens what we pass to `promiseWaterfall` so that linkAssets runs at the end of the chain *always* - we will not copy assets if linking fails.
- extracts isInstalled that we can now share with eg unregister. 

Todo:
- [x] unit tests isInstalledIOS
- [X] unit tests isInstalledAndroid

After merging:
- Rewrite `unlink` part to follow the same flow as `link` - check for isInstalled before. Once that's done, remove `isInstalled` checks from unregisterNativeModule methods.

It also allows us to easily add polling for params in iOS when we make it. I would love to change config.params to config.android.params here as well.

Waiting for the feedback. All in all I think that's a good and more explicit change, `registerModule` now does not have hidden checks for `isInstalled` - all happens before executing.


grabbou:
Also @Kureev - params do not work on patches lower than 0.20 (probably because we do not find & replace them). Not sure if something we need to do though.

grabbou:
I am going to close this one and send yet another one to filter const dependencies with extracted `isInstalled` check.

grabbou:
Diff has been cleaned up. For now `registerDependencyIOS` does not return a promise because we do not poll for params yet - this can be easily added in next release.

`registerDependencyAndroid` is now async, because when dependency is not yet linked, we will poll for params from inside it.

src/link.js
```diff
@@ -28,34 +32,60 @@ function promiseWaterfall(tasks) {
...
+      dependency.name
     );
 
-    if (didLinkAndroid) {
-      log.info(`Android module ${dependency.name} has been successfully linked`);
-    } else {
+    if (isInstalledAndroid) {
       log.info(`Android module ${dependency.name} is already linked`);
+    } else {
+      tasks.push(() => pollParams(dependency.config.params)
```
grabbou: That should be imo dependency.config.android.params as in issue

src/android/isInstalled.js
```diff
@@ -0,0 +1,9 @@
+const compose = require('lodash').flowRight;
+const readFile = require('./fs').readFile;
+
+module.exports = function isInstalled(projectConfig, name) {
+  return compose(
+    (content) => ~content.indexOf(`:${name}`),
+    readFile(projectConfig.buildGradlePath)
+  )();
+};
```
grabbou: @Kureev would you mind sending unit-tests for that function to my branch?

grabbou:
Since we now check for `isInstalled` before, that leads to better messages:
screen shot 2016-04-09 at 19 14 04

src/android/unregisterNativeModule.js
```diff
@@ -4,30 +4,32 @@ const path = require('path');
...
   /**
    * @param  {String} content Content of the Settings.gradle file
    * @return {String}         Patched content of Settings.gradle
    */
-  const cutModuleFromSettings = (name) => (content) =>
-    cut(content, `include ':${name}'\n` +
-      `project(':${name}').projectDir = ` +
-      `new File(rootProject.projectDir, '../node_modules/${name}/android')\n`
-    );
+  const cutModuleFromSettings = (content) => cut(content, settingsPatch);
```
grabbou: So the issue for settings patch was - sometimes path is android/app and that will fail to discover it. Decided to just get a settings patch and search for it which is more future proof if we add more things in the future.

src/android/unregisterNativeModule.js
```diff
@@ -4,30 +4,32 @@ const path = require('path');
...
 
   /**
    * Cut module compilation from the project build
    * @param  {String} content Content of the Build.gradle file
    * @return {String}         Patched content of Build.gradle
    */
-  const cutModuleFromBuild = (name) => (content) =>
-    cut(content, `    compile project(':${name}')\n`);
+  const cutModuleFromBuild = (content) => cut(content, buildPatch);
```
grabbou: Did the same for build.gradle `just in case`

src/android/unregisterNativeModule.js
```diff
@@ -60,15 +62,7 @@ module.exports = function unregisterNativeAndroidModule(name, dependencyConfig,
     readFile(projectConfig.mainActivityPath)
   );
 
-  /**
-   * Check if module has been installed already
-   */
-  const isInstalled = compose(
-    (content) => ~content.indexOf(getAddPackagePatch(dependencyConfig)),
```
grabbou: For some reason that was broken. Since `link` uses different check that works, I decided to use it instead (`isInstalled` method was extracted as we discussed on Skype)

src/android/unregisterNativeModule.js
```diff
@@ -4,30 +4,32 @@ const path = require('path');
...
   /**
    * Cut module compilation from the project build
    * @param  {String} content Content of the Build.gradle file
    * @return {String}         Patched content of Build.gradle
    */
-  const cutModuleFromBuild = (name) => (content) =>
-    cut(content, `    compile project(':${name}')\n`);
+  const cutModuleFromBuild = (content) => cut(content, buildPatch);
 
   const getAddPackagePatch = require(`./${prefix}/addPackagePatch`);
```
grabbou: That does not work, but didn't do it because of params (we have to do something like replace all params in a packageInstance to `s+` and construct a regexp expression `new Regexp` to use it instead)

grabbou:
Everything works, also extra tests were added.

grabbou:
In next PR I would like to extract isInstalled checks from unregisterNativeModule.

grabbou:
Commits fixing unlink (Android) were reverted because @Kureev is working on Android at the same time - feel free to take inspiration from them. Definitely needs to be fixed before we ship next release.

grabbou:
LGTM from @Kureev via FB.

PR #65, opened 2016-04-09
Fix for prelink/postlink commands not being called

### LINKED ISSUE #64: The prelink and postlink commands are not being called
This issue was uncovered in realm/realm-js#361

appden:
The `promisify` helper went too deep with the arrow functions. Added a test to ensure these commands are called in the right order.

Fixes #64


grabbou:
Thanks.

In fact, it used to work in the past, see: https://github.com/rnpm/rnpm-plugin-link/commit/4f6376a67611510ec93478e4356f70900f1c5083

@Kureev seems it's been broken by your change with `promiseWaterfall`.

PR #68, opened 2016-04-10
Extract `promiseWaterfall`

grabbou:
I would like to use it in unlink as well.

This PR extracts it and adds unit tests for two basic features.


Kureev:
As you wish

PR #69, opened 2016-04-10
Make main activity detection more solid

geof90:
Addresses https://github.com/rnpm/rnpm-plugin-link/issues/26

I completely removed the "getPrefix.js" file along with a ton of other code, because they are no longer needed, I hope you guys don't mind :p. There is now only a single "makeMainActivityPatch.js" that handles all the logic of detecting the type of `MainActivity` (whether it extends `Activity` or `ReactActivity`), and applying the appropriate patch.


grabbou:
+30 -140

I love it! Thanks for that :) Waiting for @Kureev to dbl check and I think we can merge that in!

src/android/registerNativeModule.js
```diff
@@ -2,16 +2,14 @@ const readFile = require('./fs').readFile;
 const writeFile = require('./fs').writeFile;
 const compose = require('lodash').flowRight;
 const getReactVersion = require('../getReactNativeVersion');
```
grabbou: Do we actually use that function anywhere else? If not - I think we can remove it + remove `semver` dependency :)

Kureev:
I'm busy with refactoring patch system for Android right now. I think after that we can try to merge it in (but it'll require some additional efforts)

grabbou:
Isn't this something you could try merging into your branch? (would be cool if you could submit some changes so we can check as you work!)

This looks like a small improvement and since it removes ~140 lines, it could save you a bit of time.

Kureev:
@grabbou this wouldn't scale: https://github.com/rnpm/rnpm-plugin-link/pull/69/files#diff-9be25dfa687d08b7f584875b08c5b1feR11

PR #70, opened 2016-04-11
Chore/rework android patches

Kureev:


src/android/isInstalled.js
```diff
@@ -1,9 +1,8 @@
-const compose = require('lodash').flowRight;
```
Kureev: Remove `compose` where possible

src/android/patches/0.20/addPackagePatch.js
```diff
@@ -1,2 +0,0 @@
-module.exports = (config) =>
```
Kureev: We don't have an addPackagePatch anymore. We use importPatch and packagePatch instead.

src/android/patches/0.20/makeImportPatch.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function makeImportPatch(packageImportPath) {
+  return {
+    pattern: 'import com.facebook.react.ReactActivity;',
+    patch: '\n' + packageImportPath,
+  };
```
Kureev: On Android, we have an interface for the patch object: pattern + patch (both strings)

src/android/patches/0.20/makeMainActivityPatch.js
```diff
@@ -1,34 +0,0 @@
-const addPackagePatch = require('./addPackagePatch');
```
Kureev: This file has been replaced by 2 different patches

src/android/patches/applyPatch.js
```diff
@@ -0,0 +1,8 @@
+const fs = require('fs');
+
+module.exports = function applyPatch(file, patch) {
+  fs.writeFileSync(file, fs
+    .readFileSync(file, 'utf8')
+    .replace(patch.pattern, `${patch.pattern}${patch.patch}`)
+  );
```
Kureev: read -> replace -> save

src/android/patches/makeSettingsPatch.js
```diff
@@ -1,30 +1,26 @@
 const path = require('path');
+const isWin = process.platform === 'win32';
 
-module.exports = function makeSettingsPatch(name, androidConfig, params, projectConfig) {
-  const relativeSourceDir = path.relative(
+module.exports = function makeSettingsPatch(name, androidConfig, projectConfig) {
+  var projectDir = path.relative(
```
Kureev: Better naming since we use projectDir property

src/android/patches/makeStringsPatch.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function makeStringsPatch(params) {
+  const patch = Object.keys(params).map(key =>
+    `    <string moduleConfig="true" name="${key}">${params[key]}</string>`
```
Kureev: New one. Patches strings.xml (works really nice along with unlink)

src/android/patches/revokePatch.js
```diff
@@ -0,0 +1,8 @@
+const fs = require('fs');
+
+module.exports = function revokePatch(file, patch) {
+  fs.writeFileSync(file, fs
+    .readFileSync(file, 'utf8')
+    .replace(patch.patch, '')
+  );
```
Kureev: used by unlink. The opposite operation to applyPatch

src/android/registerNativeModule.js
```diff
@@ -1,36 +1,46 @@
...
   );
 
-  const performMainActivityPatch = applyPatch(
+  applyPatch(
     projectConfig.mainActivityPath,
-    makeMainActivityPatch(androidConfig, params)
+    makeImportPatch(androidConfig.packageImportPath)
   );
 
-  compose(
```
Kureev: By using applyPatch I managed to reduce a boilerplate for patch

src/android/unregisterNativeModule.js
```diff
@@ -1,75 +1,52 @@
-const readFile = require('./fs').readFile;
-const writeFile = require('./fs').writeFile;
-const path = require('path');
-const compose = require('lodash').flowRight;
+const fs = require('fs');
```
Kureev: This one is the same as link, but with some additional things:

src/android/unregisterNativeModule.js
```diff
@@ -1,75 +1,52 @@
...
-    cutModuleFromSettings(name),
-    readFile(projectConfig.settingsGradlePath)
+  const buildPatch = makeBuildPatch(name);
+  const prefix = getPrefix(getReactVersion(projectConfig.folder));
+  const makeImportPatch = require(`./${prefix}/makeImportPatch`);
+  const makePackagePatch = require(`./${prefix}/makePackagePatch`);
+  const strings = fs.readFileSync(projectConfig.stringsPath, 'utf8');
+  var params = {};
+
+  strings.replace(
```
Kureev: Composing `params` based on the strings.xml we have

src/android/patches/makeStringsPatch.js
grabbou: I think we should escape characters here, as obviously I am not sure how `my-awesome-key` is going to be handled in a Java code.

grabbou: Also, is it possible to transparently prefix `name` with `dependencyName` ? Name clashes are inventible I am afraid.

grabbou: BTW - I just double checked Android docs and you can create as many `xml` files in `res/values` as you wish. Maybe we could create a `rnpm.xml` that we fully manage? 

> Because each resource is defined with its own XML element, you can name the file whatever you want and place different resource types in one file. However, for clarity, you might want to place unique resource types in different files

>  For a file in this directory, each child of the <resources> element defines a single resource. For example, a <string> element creates an R.string resource and a <color> element creates an R.color resource.

grabbou: Also we can define other, non string parameters (that should be easier with `xmldoc` as I proposed in previous comment), see http://developer.android.com/guide/topics/resources/more-resources.html

So `params` array could accept a type - this is mainly useful when you want to store a number instead of a String - not sure if there's any use-case for that, but good to know.

Kureev: > I think we should escape characters here, as obviously I am not sure how my-awesome-key is going to be handled in a Java code (as a name)

Valid point. But I don't know any programming language that allows you to use dashes in the name. So I think we shouldn't do any premature optimizations until we have a reason for it.

> Also, rnpm already uses xml reader for android-manifest, so I think we could use it instead to avoid string parsing - https://github.com/rnpm/rnpm/blob/master/package.json#L47

Meh, we can do it, but I think it's +1 dependency and I don't see any benefits of having it there. Probably it'll remove my regexp parsing code, but what's the point?

> However - if we go with separate rnpm.xml that we fully manage, that should not be the issue since we create file, update it and delete when neccessary.

What's the problem with default strings.xml? You was against any rnpm-specific files, so why now? Which issues do you see using strings.xml?

> Also we can define other, non string parameters (that should be easier with xmldoc as I proposed in previous comment).

We don't have any use-cases so far. It that would be a case, I'll replace my implementation by `xmldoc`.

grabbou: > Meh, we can do it, but I think it's +1 dependency

It's not a +1 dependency since it's already in a tree. Just to recall what you said for lodash.

> What's the problem with default strings.xml?

Did I say there's any problem with that? To me it makes more sense to have a separate file that we fully manage instead of adding custom extraneous attribute `moduleConfig=""` that does not tell nothing.

Kureev: > > Meh, we can do it, but I think it's +1 dependency

> It's not a +1 dependency since it's already in a tree. Just to recall what you said for lodash.

I think my main concern is not about having one more dependency, but about changing the current working approach by something else without any particular reason.

> To me it makes more sense to have a separate file that we fully manage instead of adding custom extraneous attribute moduleConfig="" that does not tell nothing

It shouldn't tell anything. It's a system parameter that marks strings, added by rnpm. 
Why don't you want to use something that we already have?

grabbou: > Which issues would we have if we use strings.xml? Which benefits rnpm.xml brings to the table?

Please read my previous comments, I am not going to keep that discussion all over again ;)

grabbou: > Why don't you want to use something that we already have (strings.xml)? 

I am not against it as (if you remember), I suggested it initially in one of the issues. `rnpm.xml` was just an idea I had in my mind as a drop-in replacement for `moduleConfig` (which in fact should be named `rnpmConfig` to be more explicit about it source). 

> Which issues would we have if we use strings.xml? Which benefits rnpm.xml brings to the table?

Please read carefully what I wrote in previous comments. Apart from that, when we bring more types to the table (and I think bools are quite cool and I seen them in few modules), if you go with `strings.xml`, you are most likely to create `bools.xml`, `arrays.xml` and other file types (which is actually cool, because it follows `android` conventions) - however, in order to get all keys managed by `rnpm`, you will have to actually read, parse and concat arrays from multiple files. It's mainly a design decision regarding how we see extra parameters working in the future.

> It's a system parameter that mark strings

I don't think neither of these are correct, but I get what you mean.

Kureev: > I don't think neither of these are correct, but I am taking you just meant managed by rnpm

Yeah :)

> This is technically speaking OK because most Android apps have strings.xml, bools.xml, arrays.xml and so on - however, that means if in the future, we decide to add e.g. <bool /> - we will have to either put it in strings.xml (breaks convention) or to put it in bools.xml (makes one more file we have to load, parse and extract rnpm variables from). Having a single file just makes it easier to manage them all in one place, however at this point I understand there might be no obvious way for having it.

I totally agree with this. I think at some point (maybe even very soon) we'll have "types" for our questions and we'll be able to group answers by type (bools, strings, numbers etc) if that's what Google recommends for Android development.

> I am not against it as (if you remember), I suggested it initially in one of the issues. rnpm.xml was just an idea I had in my mind as a drop-in replacement for moduleConfig (which in fact should be named rnpmConfig to be more explicit about its source).

Yeah, that's why I used it, actually :) I think if the approach with splitting vars by type wouldn't work, I'd say we need to go with rnpm.xml approach, but I want to give it a chance first.

Kureev:
@grabbou ? :)

src/android/registerNativeModule.js
```diff
@@ -1,36 +1,46 @@
...
 
-module.exports = function registerNativeAndroidModule(name, androidConfig, params, projectConfig) {
+module.exports = function registerNativeAndroidModule(
+  name,
+  androidConfig,
+  params,
+  projectConfig
+) {
+
+  if (isInstalled(projectConfig, name)) {
```
grabbou: Current `master` does isInstalled check on its own - you don't have to check it nor return true/false (we ignore return value).

src/android/unregisterNativeModule.js
```diff
@@ -1,75 +1,52 @@
...
-   * Make a MainActivity.java program patcher
-   * @param  {String}   importPath Import path, e.g. com.oblador.vectoricons.VectorIconsPackage;
-   * @param  {String}   instance   Code to instance a package, e.g. new VectorIconsPackage();
-   * @return {Function}            Patcher function
-   */
-  const makeMainActivityPatcher = (content) => {
-    const patched = cut(content, dependencyConfig.packageImportPath + '\n');
-    return cut(patched, getAddPackagePatch(dependencyConfig));
-  };
+  if (!isInstalled(projectConfig, name)) {
```
grabbou: Leave it here before we do the same for `unlink` as a follow up

src/android/unregisterNativeModule.js
```diff
@@ -1,75 +1,52 @@
...
-   */
-  const cutModuleFromBuild = (name) => (content) =>
-    cut(content, `    compile project(':${name}')\n`);
-
-  const getAddPackagePatch = require(`./${prefix}/addPackagePatch`);
+module.exports = function unregisterNativeAndroidModule(
+  name,
+  androidConfig,
+  projectConfig
+) {
```
grabbou: Previously that fit into a single line quite well?

src/android/patches/makeStringsPatch.js
grabbou: > Yeah, that's why I used it, actually :) I think if the approach with splitting vars by type wouldn't work, I'd say we need to go with rnpm.xml approach, but I want to give it a chance first.

I still believe it might be easier to go with a separate file, otherwise it will be tricky to update (you will have to either support backwards `strings.xml`, write a migration or bump a major). That sounds like a 5 minutes tweak that will be more future proof.

src/android/patches/0.18/makePackagePatch.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function makePackagePatch(packageInstance, params) {
+  const processedInstance = packageInstance.replace(
+    /\$\{(\w+)\}/g,
+    (pattern, paramName) => params[paramName]
+      ? `this.getResources().getString(R.strings.${paramName})`
+      : null
+  );
```
grabbou: I don't like the fact this is defined 3x in patches since it does not really change.

grabbou:
```
What is your CodePush deployment key for Android (hit <ENTER> to ignore)
rnpm-link info Linking react-native-code-push android dependency
rnpm-link ERR! It seems something went wrong while linking. Error: path must be a string
Please file an issue here: https://github.com/rnpm/rnpm/issues
```

grabbou:
```
(master)⚡ [1] % rnpm unlink react-native-code-push                                                /tmp/RNTestProject
rnpm-link info Unlinking react-native-code-push android dependency
fs.js:584
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

TypeError: path must be a string
    at TypeError (native)
    at Object.fs.openSync (fs.js:584:18)
    at Object.fs.readFileSync (fs.js:431:33)
    at unregisterNativeAndroidModule (/Users/grabbou/Repositories/ExplosiveTitles/rnpm/rnpm-plugin-link/src/android/unregisterNativeModule.js:25:22)
```

grabbou:
That's how rnpm leaves `strings.xml` after playing around with it:
```xml
<resources>    <string name="app_name">RNTestProject</string></resources>
```
We have to either make sure we do not break user formatting or go with the file we manage ourselves so one cares :)

grabbou:
Also, after providing a non-empty answer to parameters, the following line does not get stripped out from MainActivity
```java
new CodePush(this.getResources().getString(R.strings.androidDeploymentKey), this, BuildConfig.DEBUG)
```

grabbou:
Another idea I suggested that apparently got lost in our discussion was -> instead of `R.strings.androidDeploymentKey` it should be `R.strings.reactNativeCodePushAndroidDeploymentKey` (I know it's ugly but meh, ideas welcome) to avoid conflicts with other libraries that may define the same parameter name.

src/android/patches/makeStringsPatch.js
Kureev: But would you specify something which is not a string or number there? I don't think it would be a case to setup object. I'm not even sure that it's possible to specify something except strings in the plist file.

I just don't want to create a custom file for rnpm, to be honest.

src/android/registerNativeModule.js
Kureev: I'm not sure what you mean by this. If module is installed already - we do nothing

src/android/unregisterNativeModule.js
Kureev: unlink is already in this PR

src/android/unregisterNativeModule.js
Kureev: Does it blocks shipping from your perspective? Or it's a nice to have?

src/android/patches/0.18/makePackagePatch.js
Kureev: We can move it out, indeed

src/android/registerNativeModule.js
grabbou: I am saying that this check is done inside `link.js` on this line (https://github.com/rnpm/rnpm-plugin-link/blob/master/src/link.js#L35) and your implementation should follow what's currently in master (no checks, no return value https://github.com/rnpm/rnpm-plugin-link/blob/master/src/android/registerNativeModule.js#L9-L36).

src/android/unregisterNativeModule.js
grabbou: That was a follow up to this comment https://github.com/rnpm/rnpm-plugin-link/pull/70#discussion_r60419334. Basically `unlink` is not yet refactored and this check is OK. I haven't made it yet mainly because I was waiting for your branch (this branch)

src/android/unregisterNativeModule.js
grabbou: I would like to keep the code consistent in terms of style as well and I consider this as a part of PR review as well. No-one is going to die if we spent on it few hours more ;)

Kureev: I agree that style is a part of the PR. Do you think this code violates our CS? I think the main idea is to keep our code lines < 80 chars. I took this approach for splitting up a function definition form the facebook, actually :)

grabbou: As you can see, it was just a question, not even a requirement. I just asked why you changed it as previously it was a single line. If that was on purpose, I just wanted to know **why**

src/android/patches/makeStringsPatch.js
grabbou: I generally agree it might be hard to see the problem / solution at this stage, so let's just leave this for now as is and maybe come back in the future once we have more background!

You win 💃

grabbou:
Do you think @Kureev it's realistic to get it ready later today / tmrw morning?

Kureev:
@grabbou: I'll do my best to try to round off it today. I'm in Barcelona now, but I have some time for rnpm as well

test/android/patches/applyPatch.js
```diff
@@ -0,0 +1,11 @@
+const chai = require('chai');
+const expect = chai.expect;
+const applyParams = require('../../../src/android/patches/applyParams');
+
+describe('applyParams', () => {
+  it('apply params to the string', () => {
+    expect(
```
grabbou: I would also `leaves null when value not provided` :)

grabbou:
Awesome, I'll be ready for merging tomorrow morning, time to go home 🎱

test/android/patches/applyPatch.js
Kureev: Fixed

Kureev:
> Another idea I suggested that apparently got lost in our discussion was -> instead of R.strings.androidDeploymentKey it should be R.strings.reactNativeCodePushAndroidDeploymentKey (I know it's ugly but meh, ideas welcome) to avoid conflicts with other libraries that may define the same parameter name.

Solved

Kureev:
> Also, after providing a non-empty answer to parameters, the following line does not get stripped out from MainActivity
```js
new CodePush(this.getResources().getString(R.strings.androidDeploymentKey), this, BuildConfig.DEBUG)
```

Solved

Kureev:
> That's how rnpm leaves strings.xml after playing around with it:
```xml
<resources>    <string name="app_name">RNTestProject</string></resources>
```
We have to either make sure we do not break user formatting or go with the file we manage ourselves so one cares :)

I can't reproduce it :'( Can you please verify that it still happens in the latest branch?

src/android/unregisterNativeModule.js
Kureev: I changed it because it was > 80 chars, so I to fit our styleguide, I moved some parts to the next strings.

grabbou:
Will do, thanks @Kureev. Would you mind checking why CircleCi is failing?

Kureev:
Oh, I changed naming of parameters a bit. Before it was `reactNativeVectorIconsSome`, now it is `reactNativeVectorIcons_some`. There are some benefits if we keep naming of the vars untouched.

grabbou:
Yeah, I agree

Kureev:
I'll push a fix for tests pretty soon

grabbou:
Cool, I'll be around later today to potentially merge that one and do some post-merging tests. Need to add that e2e suite that we do before releasing 💨

Kureev:
Tests has been fixed 👌

grabbou:
Cool, thanks. That was hopefully the last part of comments. I will get back to you once I review it.

grabbou:
FYI played around with this today and generally looks pretty well! I don't have any things to add here - I'll go through it once again in the morning.

CC: @listintangent @geof90 - any chances you could also take a look / give it a go locally? I imagine the setup is not the best as you have to clone both repos & npm link them all together, but would be awesome to make sure we don't break anything but improve

Kureev:
Thanks for reviewing it, @grabbou. Hope we'll manage to merge it in tomorrow. Fingers crossed :)

grabbou:
"Just few nitpicks" :trollface:

src/android/unregisterNativeModule.js
```diff
@@ -1,75 +1,49 @@
...
-    writeFile(projectConfig.settingsGradlePath),
-    cutModuleFromSettings(name),
-    readFile(projectConfig.settingsGradlePath)
+  const makeImportPatch = require(`./${prefix}/makeImportPatch`);
+  const makePackagePatch = require(`./${prefix}/makePackagePatch`);
+  const strings = fs.readFileSync(projectConfig.stringsPath, 'utf8');
+  var params = {};
+
+  strings.replace(
+    /moduleConfig="true" name="(\w+)">(\w+)</g,
```
grabbou: \w+ will not work with `UA-31203912` as a value. Since that's string, I think `(.*)` is fine.

src/android/unregisterNativeModule.js
```diff
@@ -1,75 +1,49 @@
...
+    (_, param, value) => {
+      params[param.slice(toCamelCase(name).length + 1)] = value;
```
grabbou: That + 1 to length is quite hard to follow. Maybe we could comment it or make a `normalisePackageName` function? Not a PR breaker, but once you are on the previous comment, that would be also nice to do.

src/android/patches/applyParams.js
```diff
@@ -0,0 +1,14 @@
+const toCamelCase = require('to-camel-case');
+
+module.exports = function applyParams(str, params, prefix) {
+  return str.replace(
+    /\$\{(\w+)\}/g,
+    (pattern, param) => {
+      const name = toCamelCase(prefix) + '_' + param;
+
+      return params[param]
+      ? `this.getResources().getString(R.strings.${name})`
```
grabbou: Nit: 2 space indent is missing here I believe

grabbou:
Not sure if bug in master or after this PR, but:
1) rnpm link react-native-code-push
2) provide empty answer
3) value is `null` (OK!), but there's still this in diff:
```
<string moduleConfig="true" name="reactNativeCodePush_androidDeploymentKey"></string>
```
Not a big deal, but we can probably do better and just avoid adding it.

grabbou:
Here's the diff after running `rnpm unlink` the above steps:
```
-<resources>
<resources>    <string moduleConfig="true" name="reactNativeCodePush_androidDeploymentKey"></string>
     <string name="app_name">hanky</string>
 </resources>
```

grabbou:
And as mentioned above, when value is e.g. `A-BC-D`, regex does not recognise it and both MainActivity & strings.xml still contains references to package.

That said @Kureev - I am not sure which ones you want to fix as a part of this PR and which not. Apart from that, it works and code-wise all is also good.

I haven't tested on lower versions than 0.24, so not sure how the conditional patches are being applied, but #promote#upgrading

PR #71, opened 2016-04-12
Fixes https://github.com/rnpm/rnpm/issues/141

grabbou:


grabbou:
Tests failing because of eslint, need to edit from Mac.

src/link.js
```diff
@@ -36,8 +36,11 @@ const linkDependencyAndroid = (androidProject, dependency) => {
     log.info(`Android module ${dependency.name} is already linked`);
     return null;
   }
+  
+  // Legacy support for top-level params that is Android-only
```
Kureev: I though the idea is that dependency.config.params would be a "common" params for both platforms. Why you call it a legacy for Android?

grabbou: Because some of the developers already defined `params` top-level. Since they are currently Android-only (more specifically, applied to Android only because iOS is not yet made), I assume they wanted it for Android. It's a legacy so that works as is after we move to platform-specific settings and can be removed later after all of them upgrade to `config.android.params`

Kureev:
Is it a still relevant PR?

grabbou:
YES, https://github.com/rnpm/rnpm/issues/141#issuecomment-208898096. Afaik it's done, we just need to remove the `legacy` comment.

grabbou:
I am going to update it soon(ish) with the iOS params being persisted to `Info.plist`. Since iOS is easy itself, we don't have to do anything but polling :)

PR #79, opened 2016-05-02
Update `unlink` interface

grabbou:
Scope:
1) Remove `return` value from `registerDependencyAndroid` as these functions should not have return value
2) Remove `isInstalled` checks and return values from `unregister` versions as these functions should not have return value and the check is done separately
3) Add `isInstalled` check inside `unlink`
4) Fix scope in try-catch block for ES6 compatibility 

This PR is important to ship as it looks like https://github.com/rnpm/rnpm-plugin-link/pull/70/files#diff-7c8e252a3de28dabaaf67dd7fdf34a29L64 already did point 2 for `android`. That means on Android, we do not check if dependency is installed before linking.

### DIFF
--- a/src/unlink.js
+++ b/src/unlink.js
@@ -23,7 +23,7 @@ const unlinkDependencyAndroid = (androidProject, dependency, packageName) => {
   const isInstalled = isInstalledAndroid(androidProject, packageName);
 
   if (!isInstalled) {
-    log.info(`Android module ${packageName} is already unlinked`);
+    log.info(`Android module ${packageName} is not installed`);
     return;
   }
 
@@ -42,7 +42,7 @@ const unlinkDependencyIOS = (iOSProject, dependency, packageName) => {
   const isInstalled = isInstalledIOS(iOSProject, dependency.ios);
 
   if (!isInstalled) {
-    log.info(`iOS module ${packageName} is already unlinked`);
+    log.info(`iOS module ${packageName} is not installed`);
     return;
   }
 
--- a/src/unlink.js
+++ b/src/unlink.js
@@ -4,6 +4,8 @@ const log = require('npmlog');
 const getProjectDependencies = require('./getProjectDependencies');
 const unregisterDependencyAndroid = require('./android/unregisterNativeModule');
 const unregisterDependencyIOS = require('./ios/unregisterNativeModule');
+const isInstalledAndroid = require('./android/isInstalled');
+const isInstalledIOS = require('./ios/isInstalled');
 const unlinkAssetsAndroid = require('./android/unlinkAssets');
 const unlinkAssetsIOS = require('./ios/unlinkAssets');
 const getDependencyConfig = require('./getDependencyConfig');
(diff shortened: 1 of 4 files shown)


src/unlink.js
```diff
@@ -13,60 +15,73 @@ const flatten = require('lodash').flatten;
...
 
+const unlinkDependencyAndroid = (androidProject, dependency, packageName) => {
+  if (!androidProject || !dependency.android) {
+    return;
+  }
+
+  const isInstalled = isInstalledAndroid(androidProject, packageName);
+
+  if (!isInstalled) {
+    log.info(`Android module ${packageName} is already unlinked`);
```
Kureev: Maybe "Android module {...} isn't installed"?

grabbou: Yeah, sounds better!

src/unlink.js
```diff
@@ -13,60 +15,73 @@ const flatten = require('lodash').flatten;
...
-
   try {
-    const dependency = config.getDependencyConfig(packageName);
+    dependency = config.getDependencyConfig(packageName);
   } catch (err) {
-    log.warn(
-      'ERRINVALIDPROJ',
-      `Project ${packageName} is not a react-native library`
-    );
+    log.warn('ERRINVALIDPROJ', `Project ${packageName} is not a react-native library`);
```
Kureev: this line violates our 80 chars/line rule

grabbou: Don't we have a linter for that? The main reason I left it that way was because I was curious if it's an error :D

Kureev: We don't have any kind of eslint-bot here :(

But it doesn't mean that we can violate our rules 😉

grabbou: What I meant was I wasn't getting any warnings locally for this particular rule. I also thought we lint on CircleCi as well.

Kureev: wow, indeed, we don't have any rule in place for that

Kureev: https://github.com/rnpm/rnpm-plugin-link/pull/82 done

grabbou: Haha cool, will merge and rebase

grabbou:
Updated.

grabbou:
@Kureev are we good here? I'll need to update with your latest changes, but that's all right.

src/unlink.js
```diff
@@ -13,60 +15,76 @@ const flatten = require('lodash').flatten;
...
-
-    if (didUnlinkAndroid) {
-      log.info(`Android module ${packageName} has been successfully unlinked`);
-    } else {
-      log.info(`Android module ${packageName} is not linked yet`);
-    }
+  try {
+    dependency = config.getDependencyConfig(packageName);
+  } catch (err) {
+    log.warn('ERRINVALIDPROJ', `Project ${packageName} is not a react-native library`);
```
Kureev: I think it's better no make it a 3-liner:
```
log.warn(
  'ERRINVALIDPROJ', 
  `Project ${packageName} is not a react-native library`
);
```

Kureev:
Double-check > 80 char cases and merge it in 👍

PR #80, opened 2016-05-02
Improve build.gradle patch

### LINKED ISSUE #75: build.gradle patching doesn't accomodate multiple dependencies statements
If an app's `build.gradle` file includes a `buildscript.dependencies` block (e.g. because they're using [Crashlytics](https://fabric.io/kits/android/crashlytics/install) which instructs you to add it), which appears before the project's `dependencies` block, then the current [patch script](https://github.com/rnpm/rnpm-plugin-link/blob/master/src/android/patches/makeBuildPatch.js#L8) will add the linked module's `compile project()` statement to it instead of the project's `dependencies` block.

```gradle
buildscript {
    dependencies { 
       // Dependency gets placed here
    }
}

dependencies {
   // Should go here
}
```

There may be other scenarios where this could be an issue, but in general, it seems like the `makeBuildPatch.js` pattern might need to be made a little more specific. We had a CodePush user run into this issue, but it seems generic to any module that is being linked to a project that has a `build.gradle` file that matches this criteria.

grabbou:
Fixes #75 

This PR originally just added a regexp pattern (see `makeBuildPatch`). It won't pick any `dependencies {` block if it's starting with a space or a tab (handles nesting).

However, I had to modify the way we apply patches.

The original code was:
```js
replace(patch.pattern, `${patch.pattern}${patch.patch}`)
```
and that was replacing a found match with a regex & patch, e.g.:
```
/[^ \t]dependencies {\n
   compile()
}
```
since we never expected a `pattern` to be regexp.

The new code passes a function, that given a match for a given pattern, can use it to return a string to replace (e.g. match + pattern). That handles regexp being used as a pattern as we are given a real match. For plain strings, it works as previously.

Regexp can be tested here http://www.regexr.com/3dbbp

### DIFF
--- a/src/android/patches/applyPatch.js
+++ b/src/android/patches/applyPatch.js
@@ -3,6 +3,6 @@ const fs = require('fs');
 module.exports = function applyPatch(file, patch) {
   fs.writeFileSync(file, fs
     .readFileSync(file, 'utf8')
-    .replace(patch.pattern, `${patch.pattern}${patch.patch}`)
+    .replace(patch.pattern, match => `${match}${patch.patch}`)
   );
 };
--- a/src/android/patches/makeBuildPatch.js
+++ b/src/android/patches/makeBuildPatch.js
@@ -1,6 +1,6 @@
 module.exports = function makeBuildPatch(name) {
   return {
-    pattern: 'dependencies {\n',
+    pattern: /[^ \t]dependencies {\n/,
     patch: `    compile project(':${name}')\n`,
   };
 };


Kureev:
I'm up for merging it in if 

```
    dependencies {
        compile project(':test')
        compile fileTree(dir: "libs", include: ["*.jar"])
        compile "com.android.support:appcompat-v7:23.0.1"
        compile "com.facebook.react:react-native:0.18.+"
    }
```
is not a valid gradle file

grabbou:
What you mean? :) not sure if the above should work. This PR will make it fail as dependencies block cannot be indented to be recognised as a top level one. Not sure if that's good though.

Kureev:
I mean if for some reason you have a broken indentation (like 1 weird space before `dependencies {`) - it'll fail detecting it.

Kureev:
I don't want to rely on indentation in the document for pattern detection. Probably, we can merge it as a hotfix, because in the 99% that would work just fine. In the meanwhile I'll work on the proper way of doing it.

grabbou:
> P.S. I label issues I'm working on as "in-progress" to avoid other people duplicate my efforts.

That's fine, I don't mind closing it if you could come up with something soon ;) This PR is just a one-liner that allows us to use regexps in other places, so I wouldn't consider it duplicating your effort, esp. as I took a different take than mentioned.

grabbou:
Another `regex` that works regardless of indent is:
```
/[^{]\n.*dependencies {/
```

Kureev:
Merge and 🚢 it!

PR #84, opened 2016-05-16
Copy non-font assets as well [iOS]

grabbou:
ref https://github.com/rnpm/rnpm/issues/143


Kureev:
Add tests and :shipit:

grabbou:
I am not really sure how to test it since we don't unit test this particular file yet. Any chances we submit a PR with tests for that separately? (we will have to cover other bits as well)

Kureev:
Oki-doki, release 👍

PR #85, opened 2016-05-25
Exclude react-native from subprojects (dependencies)

Kureev:
Backport of https://github.com/facebook/react-native/pull/7748

### DIFF
--- a/src/android/patches/makeBuildPatch.js
+++ b/src/android/patches/makeBuildPatch.js
@@ -1,6 +1,8 @@
 module.exports = function makeBuildPatch(name) {
   return {
     pattern: /[^ \t]dependencies {\n/,
-    patch: `    compile project(':${name}')\n`,
+    patch: `    compile project(':${name}') {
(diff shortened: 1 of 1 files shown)


src/android/patches/makeBuildPatch.js
```diff
@@ -1,6 +1,8 @@
 module.exports = function makeBuildPatch(name) {
   return {
     pattern: /[^ \t]dependencies {\n/,
-    patch: `    compile project(':${name}')\n`,
+    patch: `    compile project(':${name}') {
+        exclude group: 'com.facebook.react', module: 'react-native'
```
grabbou: Unlinking wise, shouldn't we handle both cases? (to be backwards compatible) Or isn't that changing anything?

Kureev: Well, actually it does affect unlinking. Hmmm...

grabbou:
Any updates on the `unlink` part? We might need to export an array of patterns to look when unlinking.

grabbou:
Up

Issue #86, opened 2016-05-29
Android param patch incorrectly references R.strings


lostintangent:
We had someone log a bug with CodePush that seems to be due to the fact that RNPM refers to ["R.strings"](https://github.com/rnpm/rnpm-plugin-link/blob/master/src/android/patches/applyParams.js#L10) instead of "R.string" when generating the code for a configured param that is referenced in a plugin's packageInstance setting.

I haven't had a chance to look into it yet, but it seems like [R.string](https://developer.android.com/reference/android/R.string.html) is in fact the correct member name.

grabbou:
True, mind sending a pull request? cc: @Kureev

Kureev:
Oops. I'm wondering why nobody reported this before 😮

lostintangent:
Yep I can send a PR later this evening when I get home. Thanks guys!

Kureev:
I've made a PR to fix this issue. Please, take a look

PR #88, opened 2016-05-29
Change R.strings to R.string

Kureev:
https://github.com/rnpm/rnpm-plugin-link/issues/86

cc @lostintangent


lostintangent:
Looks good! Thanks so much for the quick response.

Kureev:
Then we're waiting for @grabbou's 👍

PR #89, opened 2016-06-07
Support adding/removing iOS shared libraries

appden:
When removing shared libraries, only the ones not used by anyone else are removed.

The xcode module was updated to 0.8.8 for supporting `tbd` files (text-based dynamic library definitions).

This resolves rnpm/rnpm#122 and only works with rnpm/rnpm#172 merged, which provided the `sharedLibraries` array.

### DIFF
--- a/src/ios/createGroupWithMessage.js
+++ b/src/ios/createGroupWithMessage.js
@@ -1,0 +1,25 @@
+const log = require('npmlog');
+
+const createGroup = require('./createGroup');
+const getGroup = require('./getGroup');
+
+/**
+ * Given project and path of the group, it checks if a group exists at that path,
+ * and deeply creates a group for that path if its does not already exist.
+ *
+ * Returns the existing or newly created group
+ */
+module.exports = function createGroupWithMessage(project, path) {
+  var group = getGroup(project, path);
+
+  if (!group) {
+    group = createGroup(project, path);
+
+    log.warn(
+      'ERRGROUP',
+      `Group '${path}' does not exist in your XCode project. We have created it automatically for you.`
+    );
+  }
+
+  return group;
+};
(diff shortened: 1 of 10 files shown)


Kureev:
@grabbou, can you take a look pls?

appden:
FYI, this PR has a very minor conflict with the other PR of mine (#90) that was merged. It's a very obvious one to resolve, but if you prefer, I can rebase this branch on top of what's on `master` now.

src/ios/createGroupWithMessage.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Returns the existing or newly created group
+ */
+module.exports = function createGroupWithMessage(project, path) {
+  var group = getGroup(project, path);
+
+  if (!group) {
+    group = createGroup(project, path);
+
+    log.warn(
```
grabbou: Very nice idea!

grabbou:
@appden no worries, we can rebase before merging! I have no nits here, looks very good. Great work!

I'll check it tomorrow (func wise) and see how's that working.

Do you think it's good to also test the `addGroupWithMessage` just for the sake of test coverage? The biggest issue with `link` function itself is that we haven't tested it, so now that we have moved part of it to a sep. file it might be a good idea!

appden:
@grabbou I personally really don't think so since it simply composes `getGroup` and `createGroup`, both  of which are tested.

Kureev:
So, I think we need to push this one forward :)

grabbou:
As I already answered, I'll merge it tomorrow :)

grabbou:
Merged, I've also added one commit that handles missing Frameworks group during `unlink` that I've run into while testing on a blank app. 

Apart from that - great, all builds and works nice! Thanks!

grabbou:
Took me a while, but had to rebase merge commit and change the author to @appden since the suggested merge & rebase by `Git` completely ignores the author :) Now all good.

PR #90, opened 2016-06-07
Fix bug where wrong assets are removed in unlink

appden:
The assets that should be kept were not properly specified. Also added a test for unlink that includes testing removing only the assets not used by other dependencies or the project itself.

### DIFF
--- a/src/unlink.js
+++ b/src/unlink.js
@@ -89,7 +89,7 @@ module.exports = function unlink(config, args) {
   unlinkDependencyIOS(project.ios, dependency, packageName);
 
   const allDependencies = getDependencyConfig(config, getProjectDependencies());
-  const otherDependencies = filter(allDependencies, ({name}) => name !== packageName);
+  const otherDependencies = filter(allDependencies, d => d.name !== packageName);
 
   const assets = difference(
     dependency.assets,
--- a/src/unlink.js
+++ b/src/unlink.js
@@ -10,8 +10,8 @@ const unlinkAssetsAndroid = require('./android/unlinkAssets');
 const unlinkAssetsIOS = require('./ios/unlinkAssets');
 const getDependencyConfig = require('./getDependencyConfig');
 const difference = require('lodash').difference;
+const filter = require('lodash').filter;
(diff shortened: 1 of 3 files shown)


src/unlink.js
```diff
@@ -89,10 +89,14 @@ module.exports = function unlink(config, args) {
   unlinkDependencyIOS(project.ios, dependency, packageName);
 
   const allDependencies = getDependencyConfig(config, getProjectDependencies());
+  const otherDependencies = filter(allDependencies, ({name}) => name !== packageName);
```
Kureev: I'm not sure if I understand what is `otherDependencies`. Can you explain me an idea behind it?

test/unlink.spec.js
```diff
@@ -0,0 +1,120 @@
...
+      expect(unlinkAssets.firstCall.args[0]).to.deep.equal(['Fonts/FontA.ttf']);
+      done();
+    }).catch(done);
+  });
+
+  afterEach(() => {
+    mock.stopAll();
+  });
+
+});
```
Kureev: Thanks for improving our test coverage ❤️

src/unlink.js
appden: `allDependencies` may include the one being unlinked if it's still listed in the `package.json`

Kureev: Nice. The only thing is that we're officially support node@4.1+, so we can't use ES6 destructuring 😢

appden: Ok makes sense. I will fix that. The `eslintrc` allows it, so should I remove it from there as well so nobody else gets caught by this?

Kureev: Yeah, please

appden: I pushed the change to remove destructuring. However, I didn't change the `eslintrc` since it looks like it needs a serious overall given it's using `babel-eslint` which doesn't seem to be affected by me changing `ecmaFeatures`. I'll file an issue about this.

Also, there's a test failure: `makeSettingsPatch should build a patch function`

Kureev:
Thanks <3

grabbou:
Great!

PR #100, opened 2016-09-29
`include 'app'\n` is not a necessary string

yarivkenan:
For example in f8app it does not appear.
instead there is `include ':app', ':react-native-linear-gradient'`
Some other projects might do something like `include ':react-native-send-intent', ':app'`


grabbou:
Hey,

Please submit PRs against react-native repo where `rnpm` now lives. These are frozen.

Thank you!

yarivkenan:
Thanks

